PR #6, opened 2015-09-23
Combined a bunch of semi-duplicate templates

briehl:
There was a bit of forking logic in the code to determine what templates should be used for full-featured example modules vs. simple skeletons. I moved that into the templates themselves and removed redundant ones.


briehl:
Did some more work this afternoon - the other commits add a bunch of new templates including
* README.md for most subdirectories
* stub templates for client test scripts
* Dockerfile and entrypoint.sh
* KBase license
* A semi-universal server start script (just picks one based on the chosen language and calls it start_server.sh

PR #27, opened 2015-10-28
use readlink to get paths at runtime

kkellerlbl:
Please do not merge yet, untested!  It would be helpful for others to read and try out though.

The paths to scripts and libraries should be computed at runtime, since the make may be run on a different host than the container.  Use 

	script_dir=$(dirname "$(readlink -f "$0")")'

to get the cwd of the script.


kkellerlbl:
Okay, I think this is ready now.  I've tested running make inside a new directory, and it looks okay, at least for Perl.  I haven't actually tested running the scripts yet (because I need them to run in a docker container and haven't reached that point yet).

It'd be helpful for someone to look at the Python and Java sections and see if they look good.  If so I think it's okay to merge to develop.

rsutormin:
Keith, can I make these changes as part of my current change in Makefile related to "make test" mode? In this way this PR should be closed unmerged since my PR will cover that too.

kkellerlbl:
I think that's fine.  Thanks!

PR #29, opened 2015-10-29
Docs changes

kkellerlbl:
Various docs changes for doing local deployments and testing.  Github is saying there's a conflict, but I'm not sure where.  I'll go through part of the manual merge and see what I can find.

### DIFF
--- a/doc/Module_builder.md
+++ b/doc/Module_builder.md
@@ -107,12 +107,14 @@ Since functionality in KBase is pulled into KBase from public GitHub repositorie
     git commit -m 'initial commit'
 
 Now you can sync this to a new GitHub repository. First create a new GitHub repository on github.com
-**[in the kbase project or anywhere?]**,
-but do not initialize it! Just go here to set up a new empty repository: https://github.com/new or see more instructions here: https://help.github.com/articles/creating-a-new-repository
+(it can be in your personal GitHub account or in an organization, but it must be public),
+but do not initialize it! Just go here to set up a new empty repository: https://github.com/new or see more
+instructions here: https://help.github.com/articles/creating-a-new-repository .  You may wish to
+use the name of your module as the name for your new repository.
 
 Once the repository is set up, you can push your local KBase module to GitHub.
 
-    git remote add origin git@github.com:[GITHUB_USER_NAME]/MyModule.git
+    git remote add origin git@github.com:[GITHUB_USER_OR_ORG_NAME]/[GITHUB_MODULE_NAME].git
     git push -u origin master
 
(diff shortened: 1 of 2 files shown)


kkellerlbl:
The conflict was extremely minor.  Let me see if I can fix the PR to clear it.

kkellerlbl:
My try didn't work.  Can someone review my changes and either merge by hand, or let me know they're okay and I will merge by hand?

doc/Module_builder.md
```diff
@@ -13,13 +13,14 @@ An example KBase service written in Java will be available soon.
...
-1. MacOS 8 or higher.
+1. Mac: OS X 10.8 or higher.
  - The SDK should also work on Linux but has not yet been extensively tested. We are not yet supporting Windows.
 2. Java JDK (version 7 or above): http://java.oracle.com/
  - You will also need to set the JAVA_HOME environment variable to point to your Java installation.
 3. Latest version of XCode (includes make and other developer tools)
-4. ant: http://java.oracle.com/
-5. git (command-line version): https://git-scm.com/
+4. ant: http://ant.apache.org/
+ - You will also need to add ant to your PATH environment variable.
```
nlharris: "You will also need to add ant to your PATH environment variable." -- this is almost right. You need to add to your PATH the *directory* that the ant executable lives in. I had put this instruction later:
> Also, be sure your PATH includes the directory that has the ant executable.

(Thanks for catching the mis-paste of the ant URL, though! Oops!)

doc/Module_builder.md
```diff
@@ -13,13 +13,14 @@ An example KBase service written in Java will be available soon.
...
+5. git (command-line version): https://git-scm.com/ (should come with Xcode)
```
nlharris: This sounds like you're saying that if you get git, you also get Xcode, when in fact it's the opposite. Maybe say something like "Install Xcode (URL) to get this"?

nlharris:
Your changes look fine. I don't see where the conflicts are. Is there a view that shows the conflicts?

kkellerlbl:
Thanks Nomi.  I just added your suggestions to the branch.

Unfortunately the actual conflicts are not obvious in the side-by-side view; at least, I haven't found them to be.  I've had to go to the command line to see the actual conflicts.  In this case I think the "conflict" is actually bogus but I think the only way to get around it is to use the command line to do the merge.  I can do that if my latest changes look okay.

kkellerlbl:
I merged my changes by hand.  The ''conflict'' was some extra text I added that really shouldn't have resulted in a conflict.

PR #33, opened 2015-11-04
Adding cluster.ini and ssl

sjyoo:
It took a while to figure this out and it would be nice somebody put this in

### DIFF
--- a/doc/Docker_deployment.md
+++ b/doc/Docker_deployment.md
@@ -24,6 +24,17 @@ Note: this has only been tested in OS X.
 9. docker pull kbase/deplbase
 
 ## Deploying a module
-
-1. docker build -t mymodulename .
-2. docker run --name mymodulename -d mymodulename
+1. generate cluster.ini and ssl
+```shell
+     git clone https://github.com/kbaseincubator/deploy_dev
+     cd deploy_dev/
+     cp site.cfg.example site.cfg
+     vi site.cfg
+     ./scripts/generate_config 
+     ./scripts/create_certs 
+     cp cluster.ini [your_module_dir]
+     cp -r ssl [your_module_dir]
+```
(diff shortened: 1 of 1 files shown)


doc/Docker_deployment.md
```diff
@@ -24,6 +24,17 @@ Note: this has only been tested in OS X.
...
 ## Deploying a module
-
-1. docker build -t mymodulename .
-2. docker run --name mymodulename -d mymodulename
+1. generate cluster.ini and ssl
+```shell
+     git clone https://github.com/kbaseincubator/deploy_dev
+     cd deploy_dev/
+     cp site.cfg.example site.cfg
+     vi site.cfg
```
nlharris: What's this vi step about? What are they changing in site.cfg?
I think this section needs more explanation.

kkellerlbl:
We're not really supporting a deploy_dev-based Docker build yet (though we certainly want to in the near future).  Ideally kb_sdk users would be able to choose between either running a full deploy_dev (where they would need to edit site.cfg to generate cluster.ini and ssl/), or a single container for their code which doesn't require cluster.ini and the ssl/ dir at all.

Shane and I have talked about how to make a suitable base image that doesn't need cluster.ini or ssl/ but we haven't gotten around to changing the kbase_base image to support this yet.

doc/Docker_deployment.md
sjyoo: The documentation is using "kbase/deplbase" instead of "kbase/depl"
and the user may have additional change on their site configuration.

Thanks,
-SJ

scanon:
I have a patch to the SDK that would help build the depl image.  So let me try to finish that.

scanon:
We've fixed this another way, so I'm going to close this.  Thanks Shinjae.

PR #35, opened 2015-11-05
Support to build sdkbase image

scanon:
This adds support to the SDK to build a base image for testing purposes.  It also changes the Dockerfile template to use kbase/sdkbase:latest instead of kbase/depl:latest.


kkellerlbl:
Looks good to me.  Any objections to merging?

PR #42, opened 2015-11-12
Local tests are now supported for R.

rsutormin:


msneddon:
Fantastic! Support for R is going to be very popular!

PR #52, opened 2015-11-25
Update README.md

mdejongh:
Have to start Docker before doing the make sdkbase

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,6 +58,7 @@ Build the SDK tool from source:
     cd kb_sdk
     make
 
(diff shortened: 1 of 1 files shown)


msneddon:
Thanks for the note.  There's a little better documentation on installing and having docker running in the detailed installation notes now (https://github.com/kbase/kb_sdk/blob/master/doc/kb_sdk_dependencies.md#docker), so I am going to close this PR.

What would be nice for us to add is a quick way to check if your system is compatible (e.g. correct versions of docker, docker is running, compatible version of java or make, warnings if the sdkbase image is out of date, etc).

PR #63, opened 2015-12-17
Fixed some typos, clarified some instructions

nlharris:
@mike, please see question on line 8.


psdehal:
@nlharris, yes, they have to do a full install of the kb_sdk to make the example. They don't necessarily have to do the "make sdkbase" to get the docker container. The example doesn't depend on that. But if they want to run kb-sdk test on the example, then that will use the sdkbase container. So they should do the full install.

msneddon:
Thanks!

PR #81, opened 2016-02-04
added kb_sdk_data_types.md to doc

dcchivian:


kkellerlbl:
The links to the ws docs doesn't work.  I don't think we have API docs on the web site any more, so I'm not sure where the best place to point people is.

kkellerlbl:
I skimmed the rest, it seems okay (I didn't run any of the test code though).

dcchivian:
It's cobbled together from several different Impl files, and then tweaked,
so it's definitely broken.  I still have to debug it, but wanted something
in there for the docs folks to work with.

As for Gavin's WS docs, not sure what the plan is there.  It's actually at
ci.kbase.us, but I don't think that should be the link either.

thanks,
- dylan

kkellerlbl:
The /services/ws/docs/ is apparently a new feature in the workspace service.  So the link doesn't work now but it should work by the DOE meeting.  I think (feel free to disagree) we can point to the ci docs for now, then change the links when it gets deployed upward.

MrCreosote:
> The /services/ws/docs/ is apparently a new feature in the workspace service. So the link doesn't work now but it should work by the DOE meeting. I think (feel free to disagree) we can point to the ci docs for now, then change the links when it gets deployed upward.

Correct. [This is how I do it](https://github.com/kbase/workspace_deluxe/blob/develop/readme.md#getting-documentation) in the workspace readme.

dcchivian:
added a few more tweaks

kkellerlbl:
Looks good enough, merging.  We can always revert if needed.

PR #92, opened 2016-02-19
Update module_method_spec_json.vm.properties

kkellerlbl:
Replace email address with URL.  (I wonder if we should just not provide an example here at all.)


kkellerlbl:
Didn't realize I still have an open PR!  I added docs for modifying the VirtualBox network settings to be able to pull images.

kkellerlbl:
Doh, my PR is still open!

New proposed changes under the templates to try to make it easier to run the container as nonroot.  This makes it easier to ensure that files under workdir/ are owned by the running user instead of as root, so that the user can remove the files later.

PR #93, opened 2016-02-23
Fix for SSL-related warnings in python server implementation methods.

rsutormin:


rsutormin:
Do you think it's still security breach if we run things in Docker container?

PR #116, opened 2016-04-22
"kb-sdk rename" command was added for renaming module name.

rsutormin:


msneddon:
I can do this later too, but we should also make a note to update our command completion file with the new 'rename' method and related command-line options.

PR #125, opened 2016-05-02
Fix python server error handling, don't hardcode JAVA_HOME

MrCreosote:
Currently the server is dumping the `__str__` representation of any
exceptions thrown by a server method (which includes the entire
stacktrace in the case of ServerErrors thrown by SDK clients) into the
message field of the error, which is meant for a end user readable
message. There's no way to present a reasonable error message to a user
with the current behavior, and neither the perl nor java servers behave
in the same way - they both provide short messages. Fix to copy the
message field of the caught exception as intended.

Also get rid of the seemingly unnecessary and highly annoying memoizing
of the JAVA_HOME envvar in the scripts.

### DIFF
--- a/build.xml
+++ b/build.xml
@@ -169,24 +169,18 @@ commit=${git.commit}</echo>
 
     <!-- Create main shell script-->
     <echo file="${bin.local}/${module_builder.sh.file}">#!/bin/bash
-export JAVA_HOME=${env.JAVA_HOME}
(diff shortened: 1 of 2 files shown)


msneddon:
Accepting now so we can test on CI, but @rsutormin please confirm that we don't actually need JAVA_HOME for some other reason.  To me it looks perfectly fine to remove, but maybe there is some reason in one of our environments for using JAVA_HOME.

PR #127, opened 2016-05-02
Fix SDK tests to run with no external dependencies

MrCreosote:
- make auth a submodule
	- add hack for config file to makefile
- add current directory to python path in test_client.py
- copy auth files from the submodule instead of from the dev_container

Other things that could be done, but aren't critial at all:
move typecomp into submodules dir
make kbapi-common a submodule rather than keeping the files in the
kb_sdk repo

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -4,3 +4,6 @@
 [submodule "submodules/jars"]
 	path = submodules/jars
 	url = https://github.com/kbase/jars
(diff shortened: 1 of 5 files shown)


MrCreosote:
Well, no external kbase dependencies anyway. You still have to install an asston of Perl modules and a few Python modules.

rsutormin:
Looks great!

PR #129, opened 2016-05-03
[WIP] Update module_makefile.vm.properties

kkellerlbl:
Please don't merge yet; we may want to put more into the report file.

Add KB_SDK_COMPILE_REPORT_FILE environment variable to produce the compile report.


rsutormin:
I'm not sure this is best way to do it. Let's add separate makefile target instead.

PR #134, opened 2016-05-08
Add get_provenance to test callback server

MrCreosote:
Returns some fake provenance for testing purposes.


MrCreosote:
Client side currently works like this:
https://github.com/MrCreosote/gaprice_SPAdes/compare/develop...MrCreosote:dev-prov?expand=1

PR #144, opened 2016-05-24
Created Python 3 templates

JamesJeffryes:
I've generated new templates for Python 3 and inspected them for errors with my IDE. I'll defer on further integration with the type compiler.


rsutormin:
@JamesJeffryes, thanks for this PR. We are going to use some parts of your PR to make current templates compatible with both Python version 2 and 3 at the same time. It seems to be better way than maintaining two separate sets of templates. Let us know if you have any objections.

JamesJeffryes:
That makes a lot more sense and I've updated the repo accordingly. Feel free to use whatever is helpful

MrCreosote:
@JamesJeffryes:

Hey James,
Just a quick update here - I've done a pretty extensive refactor of the py clients and incorporated py3 compatibility there and in the test template (see https://github.com/kbase/kb_sdk/pull/154). The server is going to be trickier because it uses gevent, and when we wrote the server several years ago we used 1.0rc2 which isn't py3 compatible. We're going to need to do some reasonably extensive testing on py2 and py3 to make sure updating gevent doesn't break anything, which means expanding our test rig to handle py3. It's not clear when we're going to have time to do this, but it's on our list.

Cheers, Gavin

rsutormin:
@JamesJeffryes:
Weird, but when I try to use your templates with python3 I've got this error:
  File "basicsrvServer.py", line 9, in <module>
    from jsonrpcbase import JSONRPCService, InvalidParamsError, KeywordError,\
  File "/Library/Frameworks/Python.framework/Versions/3.3/lib/python3.3/site-packages/jsonrpcbase.py", line 157
    except ParseError, e:
                     ^
SyntaxError: invalid syntax

rsutormin:
@JamesJeffryes, and here is my update to last state of develop branch: https://github.com/JamesJeffryes/kb_sdk/pull/1

MrCreosote:
@rsutormin
Looks like jsonrpcbase isn't py3 compatible either. Needs to be
except ParseError as e:

But we should check if there's a newer version of the lib before we override even more of it.

JamesJeffryes:
I can't find a newer version of jsonrpcbase. The best alternative package seems to be [here](https://github.com/pavlov99/json-rpc) with 2.6+ & 3.3+ support but a different interface. Looking through the existing library, in addition to  "Error as e" patch there's a few more conversions to make (basestring -> string and long -> float).

scanon:
What should happen to this?

MrCreosote:
We should do it, which entails updating gevent, finding or writing a replacement for jsonrpcbase, making the rest of the code py2/3 compatible, and extensive testing to make sure everything still works (including non-blocking io). *When* we should do it is the issue, since resources are pretty tight at the moment.

JamesJeffryes:
I’m actually on internship till Nov. so it’s not pressing for me. I’m happy to help anyway I can once I return because I’ve got a python 3 module I’d like to be able to publish. Thanks for all your help in this.

sychan:
Thanks for submitting the PR, but it looks like we'll need a more concerted effort for Python3 compatibility.

PR #154, opened 2016-05-31
Make a static python base/generic client with template generated thin wrappers

MrCreosote:
Note: this PR sits on top of https://github.com/kbase/kb_sdk/pull/152. Close that one first.
Changes in just this PR: https://github.com/kbase/kb_sdk/compare/2402114c42b9f97d683f1d7cfb80e2cd47c59ea0...279a9b64c61176b06a8df897adcde4e107865f52

also:
* fix error handling in the NJSW mock service processRpcCall method.
* add verbose mode to kb-sdk compile
* EDIT: make py test template py3 compatible per https://github.com/kbase/kb_sdk/pull/144

TODO (beyond those in PR https://github.com/kbase/kb_sdk/pull/152):
* Add dynamic init client functionality


MrCreosote:
@msneddon @rsutormin One thing we might want to discuss is where the baseclient goes. Currently every module will have a copy of the baseclient which means a user has to import ServerError from multiple base clients to handle all errors. If the baseclient is in, for example, the biokbase module, then there's only one copy assuming all other modules are inside biokbase.

PR #158, opened 2016-06-02
Fix callback server tests broken by changes to NetUtils

MrCreosote:


MrCreosote:
@rsutormin I just added a commit to bugfix authdef handling. All kb_sdk and workspace tests pass.

Ignore whitespace: https://github.com/kbase/kb_sdk/pull/158/files?w=1

src/java/us/kbase/common/utils/NetUtils.java
```diff
@@ -26,7 +26,8 @@
         List<String> ret = new ArrayList<String>();
         for (Enumeration<NetworkInterface> en = NetworkInterface.getNetworkInterfaces(); en.hasMoreElements();) {
             NetworkInterface intf = en.nextElement();
-            if (!intf.isUp()) continue;
+            // breaks linux, if required for windows needs to be os-dependent
```
sjyoo: Sorry... I think it is fine with Windows with the current code... 
My earlier code has to use this...

PR #174, opened 2016-07-08
Fix for duplicated KIDL type/function declarations. Binary jar update.

rsutormin:


rsutormin:
@MrCreosote: this should fix your problem with duplication of function declaration.

PR #177, opened 2016-07-26
Fix to worker with Docker for Mac

scanon:
The new Docker for Mac doesn't use Virtual box


rsutormin:
Works good with docker-machine on Mac as well.

PR #181, opened 2016-07-27
Changes in async call processing from clients in all languages.

rsutormin:
Now in async call check cycle we start from 100 ms waiting and then multiply it by 1.5 every step. When it reaches 5 minutes (aprox. after 16 minutes in common) we keep it on 5 min. level.


MrCreosote:
Code looks ok, but there's no tests. Actually testing the scaling completely is probably not reasonable, but should at least test the setters/getters for the languages and run some jobs that will trigger a few wait time scalings.

rsutormin:
@MrCreosote, good point but will be done after my vacation. Can we merge it now (at least I tested default scaling manually with https://github.com/kbase/kb_sdk/pull/181/files#diff-35f5ccc2722169b21af7827eb9fc68ffR347).

PR #184, opened 2016-08-02
Update python server for new auth

MrCreosote:
Also some pep8ification.
Also fixed the python SSL warnings.

* new simple auth client
* uses KBase auth server vs. globus
* auth-service-url in deploy.cfg now sets kbase auth url

TODO:
* [x] auth client tests, incl. token cache

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,4 @@
 /classes/
 /temp_*
 /ASimpleModule*
+/TestAsyncPerl/
--- a/Makefile
+++ b/Makefile
@@ -88,8 +88,9 @@ sdkbase:
 	docker build --no-cache -t kbase/kbase:sdkbase.latest sdkbase
 
 test: submodule-init
-	@# todo: remove perl typecomp tests and add it as a separate target
 	@echo "Running unit tests"
+	nose2 -s test_scripts/py_module_tests -t src/java/us/kbase/templates
+	@# todo: remove perl typecomp tests and add it as a separate target
 	$(ANT) test
 
 test-client:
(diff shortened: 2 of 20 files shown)


msneddon:
Changes look reasonable from a quick review and kb-sdk still compiles and seems to run correctly, so I wouldn't hold up on the catalog work, but a few comments.

First, we should start documenting the list of dependencies we need for tests.  I know we probably aren't doing this yet, but this is good time to start the list, maybe in some new file in the docs directory.  The two new dependencies I personally needed were: nose2 and configobj

Second- more of a question- we need to switch to tokens instead of usernames/passwords in test configs.  I put in the old globus token, and it didn't complain about that, but I assume the globus token won't work at some point.  Is that point coming soon?  The only way to get a token will be through some user profile page, correct?  This is only an issue because most of our internal developers are using the develop branch, and we need that to be relatively stable through the end of the quarter.  So more generally, if we don't think this auth related updates can go out to prod before the end of the quarter, I'm hesitant on putting it on develop.  We should maintain it in a separate feature branch.

And finally- I'm getting a test failure:
======================================================================
FAIL: test_bad_url (py_module_tests.test_auth_client.TestAuth)
----------------------------------------------------------------------
  File "/Users/msneddon/kb/sdk_work/kb_sdk/test_scripts/py_module_tests/test_auth_client.py", line 80, in test_bad_url
    "HTTPSConnectionPool(host='thisisasuperfakeurlihope.com'"))
AssertionError: False is not true

MrCreosote:
I updated the tests to not need configobj and to print out a better error for the test that failed for you. Can you try again? I also started a test dependencies list.

I'm not sure I understand your concerns about switching from uid/pwd to tokens. Let's talk in person.

MrCreosote:
for posterity, I addressed Mike's switching concerns and he is thoroughly placated.

PR #190, opened 2016-08-10
Fix related to recent changes in auth.

rsutormin:


MrCreosote:
Confused about one thing - it looks like now there's only one list of jars, and it's stored in JAR_DEPS. But might not make and make bin have different jar requirements since the prebuilt jar might be older than the source code? Or am I not getting this?

rsutormin:
Yeah, good point. There are two alternative solutions I see here. First is to have two separate jar lists like JAR_DEPS and BIN_JAR_DEPS. Second is to refresh (and recommit) jar of kb-sdk into lib/ folder every time we change JAR_DEPS (or at least we make major change affecting compiled code). For second solution we have to add some test running kb-sdk through "make bin" and making sure it's not broken (it may be even useful to it in either way). @MrCreosote, what's your preference?

MrCreosote:
#1 sounds right to me. Otherwise we'll bloat the repo with jar versions. Also if we can choose when to update the bin jar that can act as a stable dev version.

Is it possible to add comments in the jar deps files that explains what they are and when to update them?

rsutormin:
@MrCreosote: I think PR is ready for review.

Makefile
```diff
@@ -14,6 +14,8 @@ DIR = $(shell pwd)
 
 ANT ?= ant
 ANT_OPTIONS =
+KBASE_COMMON_JAR = kbase/common/kbase-common-0.0.20.jar
```
MrCreosote: What's the purpose of specifying the common jar in the makefile?

Makefile
```diff
@@ -43,16 +45,30 @@ else
 ###############################
 
 compile: jars-submodule-init
-	$(ANT) -Djardir=submodules/jars/lib/jars/
+	$(ANT) -Djardir=submodules/jars/lib/jars/ -DKBASE_COMMON_JAR=$(KBASE_COMMON_JAR)
 
 endif
```
MrCreosote: It doesn't look like there's a target for updating the prebuilt jar file. It might make sense to add a target that builds the jars, moves it to wherever it's supposed to go, and copies JAR_DEPS to BIN_JAR_DEPS adding a comment that BIN_JAR_DEPS shouldn't be manually edited. What do you think?

src/java/us/kbase/mobu/initializer/test/DockerClientServerTester.java
```diff
@@ -166,9 +168,36 @@ protected File initPython(String moduleName) throws Exception {
         return moduleDir;
     }
     
+    public static void correctDockerfile(File moduleDir) throws Exception {
+        File buildXmlFile = new File("build.xml");
+        File sdkSubFolder = new File(moduleDir, "kb_sdk");
+        File sdkDistSubFolder = new File(sdkSubFolder, "dist");
+        FileUtils.copyFile(buildXmlFile, new File(sdkSubFolder, buildXmlFile.getName()));
```
MrCreosote: Does this create directories if missing? Or does /kb/module/kb_sdk/dist already exist (I dont think so?)?

Makefile
rsutormin: The purpose is to support JENKINS tests (which we stopped to run unfortunately) where you compile temporary common jar from develop branch of java_common.

src/java/us/kbase/mobu/initializer/test/DockerClientServerTester.java
rsutormin: They do create all subfolders when you copy:
http://grepcode.com/file/repo1.maven.org/maven2/commons-io/commons-io/1.0/org/apache/commons/io/FileUtils.java#409

Makefile
rsutormin: Sounds reasonable. Will add it soon.

Makefile
```diff
@@ -105,7 +110,7 @@ sdkbase:
 
 test: submodule-init
 	@echo "Running unit tests"
-	nose2 -s test_scripts/py_module_tests -t src/java/us/kbase/templates
+	@# nose2 -s test_scripts/py_module_tests -t src/java/us/kbase/templates
```
MrCreosote: I don't think this should be commented out for everyone - it should be commented out by a particular user if they need to do so.

MrCreosote:
here's the changes I made to get the tests to pass and reenable the py tests. If you can patch these in I think this is g2g.

```
crusherofheads@icrushdeheads:~/localgit/rsutormin/kb_sdk$ git diff
diff --git a/Makefile b/Makefile
index 43d6449..7395d86 100644
--- a/Makefile
+++ b/Makefile
@@ -110,7 +110,7 @@ sdkbase:
 
 test: submodule-init
        @echo "Running unit tests"
-       @# nose2 -s test_scripts/py_module_tests -t src/java/us/kbase/templates
+       nose2 -s test_scripts/py_module_tests -t src/java/us/kbase/templates
        @# todo: remove perl typecomp tests and add it as a separate target
        $(ANT) test -DKBASE_COMMON_JAR=$(KBASE_COMMON_JAR)
 
diff --git a/build.xml b/build.xml
index 4e1dada..b27343b 100644
--- a/build.xml
+++ b/build.xml
@@ -108,7 +108,7 @@ giturl=${git.url}
 branch=${git.branch}
 commit=${git.commit}</echo>
     <echo>Compiling with jar directory: ${jardir}</echo>
-    <javac destdir="${classes}" srcdir="${src}" includeantruntime="false" debug="true" classpathref="compile.classpath" />
+    <javac destdir="${classes}" srcdir="${src}" includeantruntime="false" debug="true" classpathref="compile.classpath" target="1.7" source="1.7" />
     <!-- Copy resource files-->
     <copy todir="${classes}">
       <fileset dir="${src}">
```

Makefile
rsutormin: I'm sorry, Gavin, it was just my mistake. :(

rsutormin:
@MrCreosote: Fixed.

PR #194, opened 2016-09-20
Added travis-ci config

sychan:
Put in a minimal .travis.yml file that builds the bin directory and the sdk base container.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,10 @@
+language: java
+
+sudo: required
+
+services:
+  - docker
+
+script:
+  - make bin
+  - make sdkbase


.travis.yml
```diff
@@ -0,0 +1,10 @@
+language: java
+
+sudo: required
+
+services:
+  - docker
+
+script:
+  - make bin
```
msneddon: quick note- let's just run make.  make bin does not actually compile the sdk, it uses a prebuilt jar.

rsutormin:
@sychan: Hey Steve, do you want to merge this PR as is or you're going to work on it further?

sychan:
Lets cancel this PR, it was against the wrong branch anyway.

PR #199, opened 2016-10-05
Add requests_toolbelt to sdkbase

kkellerlbl:
Add requests_toolbelt to Dockerfile for sdkbase image.

I am running tests now.  When they pass this should be safe to merge.


kkellerlbl:
My tests passed locally.

PR #214, opened 2016-11-09
Support for local test image in callback server in "kb-sdk test" mode.

rsutormin:


src/java/us/kbase/mobu/tester/SDKSubsequentCallRunner.java
```diff
@@ -36,6 +40,43 @@ public SDKSubsequentCallRunner(
...
+        Map<String, String> localModuleNameToImage = getLocalModuleNameToImage();
+        return localModuleNameToImage != null && 
+                localModuleNameToImage.containsKey(module);
+    }
+    
+    @Override
+    protected ModuleRunVersion createModuleRunVersion(ModuleMethod modmeth,
+            String serviceVer, ModuleVersion mv) throws MalformedURLException {
+        return super.createModuleRunVersion(modmeth, serviceVer, mv);
+    }
```
MrCreosote: Doesn't seem to be any need to override this

src/java/us/kbase/mobu/tester/SDKSubsequentCallRunner.java
```diff
@@ -36,6 +40,43 @@ public SDKSubsequentCallRunner(
...
+                    .withGitUrl("http://localhost")
+                    .withGitCommitHash("local-docker-image").withVersion("local");
+        } else {
+            return super.loadModuleVersion(modmeth, serviceVer);
+        }
+    }
+
+    protected Map<String, String> getLocalModuleNameToImage() {
+        return null;
+    }
```
MrCreosote: Should note in comment that this method is intended for override to provide local module image names.

src/java/us/kbase/mobu/tester/ModuleTester.java
```diff
@@ -229,14 +230,16 @@ public void logNextLine(String line, boolean isError) {
                         }
                     }).build();
             ModuleRunVersion runver = new ModuleRunVersion(
-                    new URL("https://fakefakefakefakefake.com"),
-                    new ModuleMethod("use_set_provenance.to_set_provenance_for_tests"),
-                    "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa", "0.0.0", "dev");
+                    new URL("https://localhost"),
+                    new ModuleMethod(moduleName + ".to_set_provenance_for_tests"),
```
MrCreosote: I would remove to_set_prov... here and change to .run_local_tests or something like that.

src/java/us/kbase/mobu/tester/ModuleTester.java
```diff
@@ -229,14 +230,16 @@ public void logNextLine(String line, boolean isError) {
...
+                    "local-docker-image", "local", "dev");
             final DockerMountPoints mounts = new DockerMountPoints(
                     Paths.get("/kb/module/work"), Paths.get("tmp"));
+            Map<String, String> localModuleToImage = new LinkedHashMap<>();
+            localModuleToImage.put(moduleName, imageName);
```
MrCreosote: My personal preference would be to make a small class instead of a map that takes a ModuleMethod and the String imageName, but I'll leave that up to you.

MrCreosote:
Oh, don't forget to move the changes in the call runner to NJSW. We should make a jar out of that stuff at some point...

src/java/us/kbase/mobu/tester/ModuleTester.java
rsutormin: I don't think method is necessary in this structure. And since it all exists in purely SDK code (not common execution part) we can change it later at any time. And I'm too lazy to it now anyway.

rsutormin:
Changes are done in NJSW as well: https://github.com/kbase/njs_wrapper/pull/63

rsutormin:
@MrCreosote: Test is added.

MrCreosote:
Tests pass

PR #217, opened 2016-11-30
Update README.md

realmarcin:


msneddon:
Please create PRs against the develop branch.

Although I'm not sure this instruction should be in the quick_start.  The Quick Start assumes a new module, and in fact already tells you to edit the test_local/test.cfg file in the line right above your change.  A better place would be in the docs on testing:
https://github.com/kbase/kb_sdk/blob/develop/doc/kb_sdk_local_test_module.md

There you would be able to provide a little more info in tutorial form saying something like: if the test_local folder is not there, kb-sdk test will create it for you.  Make sure you update the test_local/test.cfg file... etc.

PR #219, opened 2016-12-02
Got rid of contact help@kbase.us in example spec.json

nlharris:
I already did this in the develop branch but no one will tell me whether commits to this repo should go into develop or master.


msneddon:
PRs should go to the develop branch.  We're going to update the master branch today I think, so if it's already in develop it will probably go out tomorrow.

nlharris:
Ok. I had asked which branch I should PR to (in JIRA and in Slack) and no one answered me.

PR #225, opened 2017-01-21
This adds support for running kb-sdk as a docker container.  Additional

scanon:
modifications are still need to the docs area.


scanon:
Hold off on merging.  I missed something.

scanon:
Once the tests complete I think this is ready for review.

msneddon:
Great!  This is really cool.  It all seemed to work for me- I was able to build an image from the dockerfile, generate the alias and works as expected, at least on my mac.  Docs were good, so I think we should merge.

PR #228, opened 2017-01-25
Updated Python example module

msneddon:
This updates the python example module to use local functions and the updated Assembly type.  Unit tests all just passed, but we still need to:
- review generated documentation
- ~~figure out what we want to do with other examples~~
- ~~(done) update tab completion script~~
- ~~(done: MsneddonContigFilter4 in next) actually register a new python example module and make sure it is runnable in CI~~
- ~~update kb_sdk version number~~

Other minor things that would be good to do now that I'm looking at the example are:
- ~~(done) figure out again how to make the right updates to eliminate the python SSL warning (the ones we had been using were broken due to an updated python module)~~
- update the license file?
- ~~(done enough) clean up the generated README file and provide more info there~~
- update the generated ui spec/yaml to better match our conventions
- add an example local function spec in the ui directory


src/java/us/kbase/mobu/initializer/ModuleInitializer.java
```diff
@@ -125,7 +124,7 @@ public void initialize(boolean example) throws Exception {
 		templateFiles.put("module_readme_ui", Paths.get(moduleDir, "ui", "README.md"));
 		templateFiles.put("module_readme_test", Paths.get(moduleDir, "test", "README.md"));
 		templateFiles.put("module_readme_docs", Paths.get(moduleDir, "docs", "README.md"));
-		templateFiles.put("module_readme_data", Paths.get(moduleDir, "data", "README.md"));
+		//templateFiles.put("module_readme_data", Paths.get(moduleDir, "data", "README.md"));
```
scanon: Why are you removing the data directory?

src/java/us/kbase/mobu/initializer/ModuleInitializer.java
```diff
@@ -219,12 +217,31 @@ public void initialize(boolean example) throws Exception {
...
+			}
+		}
+
 		System.out.println("Done! Your module is available in the " + moduleDir + " directory.");
 		if (example) {
 			System.out.println("Compile and run the example methods with the following inputs:");
 			System.out.println("  cd " + moduleDir);
-			System.out.println("  make          (could be necessary after changes in " + new File(specFile).getName() + ")");
-            System.out.println("  kb-sdk test   (will require to set GlobusOnline test account in test_local/test.cfg)");
+			System.out.println("  make          (required after making changes to " + new File(specFile).getName() + ")");
```
scanon: Is make really required since it would get run in the build?

src/java/us/kbase/templates/module_dockerfile.vm.properties
```diff
@@ -2,13 +2,24 @@ FROM kbase/kbase:sdkbase.latest
...
 
 # RUN apt-get update
 #if( $language == "python")
+
 RUN pip install coverage
+
+# update security libraries in the base image
+RUN pip install cffi --upgrade \
+    && pip install pyopenssl --upgrade \
+    && pip install ndg-httpsclient --upgrade \
```
scanon: We ran into a problem with ndg-httpsclient today.  Not sure if this resolves it.

scanon:
It looks good so far.  Few remarks inline.

src/java/us/kbase/mobu/initializer/ModuleInitializer.java
msneddon: I was thinking about removing things that aren't used, but realized this is used for ref data.  I'll leave this and the other ui/widgets placeholder directories for now.

src/java/us/kbase/mobu/initializer/ModuleInitializer.java
msneddon: If a new function is added, there's no stub in the Impl to add code until you run make- although yes, other changes wouldn't necessarily require a make.

src/java/us/kbase/templates/module_dockerfile.vm.properties
msneddon: yes- we ran into the same problem and this is a fix courtesy of Gavin.  @MrCreosote you actually don't need the first sudo apt-get install line because those things are already up to date in the base image.

MrCreosote: They're up to date for now. They won't be in the future, and my feeling was we should keep the security stuff as new as possible.

msneddon:
Ok, I think this is good enough now to do a final review and merge.

src/java/us/kbase/templates/module_dockerfile.vm.properties
msneddon: good point, although really we should be doing all of this in the base image.

PR #230, opened 2017-01-31
update java example

MrCreosote:
* [x] fix tests
* [x] update UI specs
* [x] run kb_sdk tests
* [x] create new module with init, register, and test

### DIFF
--- a/src/java/us/kbase/templates/module_dockerfile.vm.properties
+++ b/src/java/us/kbase/templates/module_dockerfile.vm.properties
@@ -5,11 +5,13 @@ MAINTAINER #if($username)${username}#{else}KBase Developer#{end}
 # In this section, you can install any system dependencies required
 # to run your App.  For instance, you could place an apt-get update or
 # install line here, a git checkout to download code, or run any other
-# installation scripts.  Here we install a python coverage tool and an
-# https library that is out of date in the base image.
+# installation scripts.
 
 # RUN apt-get update
 #if( $language == "python")
+
+# Here we install a python coverage tool and an
+# https library that is out of date in the base image.
 
 RUN pip install coverage
 
--- a/src/java/us/kbase/templates/module_dockerfile.vm.properties
+++ b/src/java/us/kbase/templates/module_dockerfile.vm.properties
@@ -23,7 +23,7 @@ RUN pip install cffi --upgrade \
     && pip install requests --upgrade \
     && pip install 'requests[security]' --upgrade
 #end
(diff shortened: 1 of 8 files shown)


MrCreosote:
Ready for review/merge

src/java/us/kbase/templates/module_dockerfile.vm.properties
```diff
@@ -21,6 +23,11 @@ RUN pip install cffi --upgrade \
     && pip install requests --upgrade \
     && pip install 'requests[security]' --upgrade
 #end
+#if( $language == "java")
```
msneddon: We should only include this line if we are creating an example module.  There should be a flag available to indicate that it is an example.

src/java/us/kbase/templates/module_java_impl.vm.properties
```diff
@@ -59,64 +72,103 @@ public class ${java_module_name}Server extends JsonServerServlet {
...
+         * object. The return object gives us the path to the file that was created.
+         */
+        System.out.println("Downloading assembly data as FASTA file.");
+        final AssemblyUtilClient assyUtil = new AssemblyUtilClient(callbackURL, authPart);
+        /* Normally this is bad practice, but the callback server (which runs on the same machine
+         * as the docker container running the method) is http only
+         * TODO Should allow the clients to not require a token, even for auth required methods,
+         * since the callback server ignores the incoming token. No need to transmit the token
+         * here.
+         */
```
msneddon: I thought the auth token wasn't required when talking to the callback server?  At least it's not in python, but the java client may be different.

MrCreosote: It isn't, the requirement is imposed by the java compiled clients. That's why the TODO is there - we need to change the clients so that we aren't forced to include the token.

src/java/us/kbase/templates/module_dockerfile.vm.properties
MrCreosote: fixed

MrCreosote:
@msneddon ready for re-review, tests passed

src/java/us/kbase/templates/module_java_impl.vm.properties
msneddon: I see, ok

msneddon:
great, looks good.  I'm running local tests, and will pull this in if it passes.  @rsutormin you should still take a pass through this as well.

PR #191, opened 2016-08-25
[WIP] Allow configuration of auth service

kkellerlbl:
Support configuration of an auth service endpoint in the configuration file.  The generated Service.pm will use the auth client library's default URL if not configured.

I've been testing this with my modifications to handle_service and it looks like it behaves okay.  It'd be helpful to have other eyes on it as well.

This PR goes to a new branch of kb_sdk, so merging won't break anything.  If it looks ok we can do another PR to develop.

PR #240, opened 2017-02-21
[WIP] changes for configurable auth server support in perl server

kkellerlbl:
Please don't merge yet.


kkellerlbl:
Last I left this code, it was still failing some tests, but I haven't run them in a while.  I'll do that myself but it'd be useful for others to check it too.

Makefile
```diff
@@ -14,7 +14,7 @@ DIR = $(shell pwd)
 
 ANT ?= ant
 ANT_OPTIONS =
-KBASE_COMMON_JAR = kbase/common/kbase-common-0.0.20.jar
+KBASE_COMMON_JAR = kbase/common/kbase-common-0.0.22.jar
```
MrCreosote: Test dependencies from https://github.com/kbase/kb_sdk/pull/191/files would be useful here

src/java/us/kbase/templates/perl_server.vm.properties
```diff
@@ -261,7 +261,14 @@ sub call_method {
...
 
-	my $auth_token = Bio::KBase::AuthToken->new(token => $token, ignore_authrc => 1);
+	my $auth_token;
+        if ($self->config->{'auth-service-url'})
+        {
+            $auth_token = Bio::KBase::AuthToken->new(token => $token, ignore_authrc => 1, auth_svc=>$self->config->{'auth-service-url'});
+        } else {
+            $auth_token = Bio::KBase::AuthToken->new(token => $token, ignore_authrc => 1);
+        }
+
```
MrCreosote: here's a relevant comment from the last PR with these changes
> I want to run three different permutations of test.

>    Current codebase
>    These changes plus stock auth (should work, I think)
>    These changes plus Perl auth client changes in auth/dev-perl_auth branch

> So far test 1. runs for a long time, so I may need to kill it and start again tomorrow. 3. is having problems, and I need to track down exactly what they are. It's currently failing in testJavaAsyncService in AsyncDockerTest, which is odd because I didn't touch that code. It may be a different issue.

So changes may be necessary to get tests to pass.

Makefile
kkellerlbl: The test_dependencies.md changes got included in another PR.

rsutormin:
@kkellerlbl: Keith, why don't we close this PR and later we'll synchronize develop with auth2 branch directly?

kkellerlbl:
Agreed

PR #243, opened 2017-02-23
First part of changes related to switch to auth2 (tests are broken for now).

rsutormin:


kkellerlbl:
Should we merge this?  It's in a separate branch so I think it'd only impact the auth2 environment.

msneddon:
Looks good to me too.

One comment is that we're starting to get a lot of config URLs.  Would be good to refactor some of this into parameterized functions so we don't have to paste in each url variable in multiple places.

PR #244, opened 2017-03-06
added exit code to runTestCommand

arfathpasha:
Added code to return a non-zero exit code when unit tests are executed.

Currently, the 0 exit code is returned even when unit tests fail. 

This affects travis test reporting when integration tests are built using a call to 'kb-sdk test'.

### DIFF
--- a/src/java/us/kbase/mobu/ModuleBuilder.java
+++ b/src/java/us/kbase/mobu/ModuleBuilder.java
@@ -343,7 +343,7 @@ public class ModuleBuilder {
     private static int runTestCommand(TestCommandArgs testArgs, JCommander jc) {
         // Figure out module name.
         // Join together spaced out names with underscores if necessary.
-        int returnCode = 131;
+        int returnCode = 1;
 
         try {
             ModuleTester tester = new ModuleTester();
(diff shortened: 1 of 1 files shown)


src/java/us/kbase/mobu/ModuleBuilder.java
```diff
@@ -343,17 +343,20 @@ private static String getCmdOutput(File workDir, String... cmd) throws Exception
     private static int runTestCommand(TestCommandArgs testArgs, JCommander jc) {
         // Figure out module name.
         // Join together spaced out names with underscores if necessary.
+        int returnCode = 131;
```
rsutormin: PR looks good except mysterious 131 exit code. This code would be default one if KB_SDK couldn't run test process at all. Like when token authentication failed. Or file system on host machine doesn't allow to put token/config files in place. What exit code reference do you mean here?

arfathpasha: 131 represents "State not recoverable" in Linux. I guess defaulting to 1 "Operation not permitted" would be another option. Anything you see fit work for me.

arfathpasha: FWIW, the system returns 2 when unit tests fail. But 2 represents "No such file or directory"...

rsutormin: Honestly "131" seems to be too extravagant. I would use "1" by default.

arfathpasha: I agree.

PR #252, opened 2017-04-07
Merging all auth2 changes into develop

rsutormin:


src/java/us/kbase/kidl/KidlParser.java
```diff
@@ -117,149 +103,4 @@
 		}
 		return ret;
 	}
-
```
rsutormin: @kkellerlbl: This is deletion of old perl type-compiler support.

PR #254, opened 2017-04-25
Synchronizing develop -> master

rsutormin:


kkellerlbl:
Looks okay to me.  One of the Travis jobs failed, is that a problem?

kkellerlbl:
I re-ran the failing job.  It looks like it was having trouble pulling from Dockerhub.

PR #257, opened 2017-07-27
Fix python auth client & tests

MrCreosote:
Looks like the auth client was never compatible with with auth2 error
syntax. No need to support auth1 at this point.

Later, update the client to use the new auth2 api rather than the legacy
api, including implementing the suggested cache time.

Tested that the default url works manually.


MrCreosote:
@arfathpasha @ugswork

PR #258, opened 2017-07-27
Synch up develop with master

MrCreosote:
Couple of commits made directly to master instead of develop


MrCreosote:
Changes are already in master sooooo

PR #261, opened 2017-07-28
Fix tests

MrCreosote:
TODO:
* [x] Fix CallbackServer tests
* [x] Always require casperjs in the path to force JS client tests


MrCreosote:
@arfathpasha @ugswork

MrCreosote:
Fixes for SDK tests required in other modules:
https://github.com/kbasetest/njs_sdk_test_1/commit/03b9210ac825858c1ae6339786686cea5dac5929
https://github.com/kbasetest/njs_sdk_test_2/commit/284948a93524ba51e48b1c71693b269647c125c3

Note that the NJSW tests also use these modules, and may check for the commits (or versions), so making these changes to the modules might break those tests.

PR #262, opened 2017-08-02
Update install docs to make docker method the default

JamesJeffryes:
Addresses PTV-427 on Jira
https://kbase-jira.atlassian.net/browse/PTV-427


msneddon:
Awesome- looks great!

PR #264, opened 2017-08-07
Sdk doc edits

JamesJeffryes:
Pursuant to PTV-537, I've made a number of changes to make the SDK docs a little clearer and more comprehensive.

* Broke up the 4th step into more equitable sized portions.
* Explained installation of KBase modules
* More explanation of UI specification
* Discussion of inputs and outputs to KBase methods

https://kbase-jira.atlassian.net/browse/PTV-537


doc/kb_sdk_register_module.md
```diff
@@ -4,17 +4,19 @@
...
+6. [Specify User Interface](kb_sdk_make_ui.md)
+7. [Locally Test Module and Method(s)](kb_sdk_local_test_module.md)
+8. **Register Module**
+9. [Test in KBase](kb_sdk_test_in_kbase.md)
+10. [Complete Module Info](kb_sdk_complete_module_info.md)
+11. [Deploy](kb_sdk_deploy.md)
 
 
-### 6. Register Module
+### 8. Register Module
```
sychan: Recently it became apparent that we may not have documented that a developer needs to register for a developer account at https://accounts.kbase.us/index.php?tpl=request_identity.tpl and make sure that their account is created before they can register a module. Can you add this into the docs in this file?

sychan:
@MrCreosote Can you review this as well? I've made my suggestions for edits.

doc/kb_sdk_impl_methods.md
```diff
@@ -0,0 +1,201 @@
...
+##### <A NAME="impl-shell-tool"></A>5.6. Invoking Shell Tool
+
+```python
+        command_line_tool_params_str = " ".join(command_line_tool_params)
+        command_line_tool_cmd_str = " ".join(command_line_tool_path, command_line_tool_params_str)
+        
+        # run <command_line_tool>, capture output as it happens
+        self.log(console, 'running <command_line_tool>:')
+        self.log(console, '    '+command_line_tool_cmd_str))
+        p = subprocess.Popen(command_line_tool_cmd_str,
```
MrCreosote: This is extremely bad practice - the command line should be provided as a list of arguments. See https://docs.python.org/2/library/subprocess.html

Example: https://github.com/kbaseapps/kb_quast/blob/master/lib/kb_quast/kb_quastImpl.py#L119-L122

doc/kb_sdk_impl_methods.md
```diff
@@ -0,0 +1,201 @@
...
+                    cwd = self.scratch,
+                    stdout = subprocess.PIPE,
```
MrCreosote: Why bother with piping things around? Just use call and things will happen as they should.

doc/kb_sdk_impl_methods.md
```diff
@@ -0,0 +1,201 @@
...
+        
+        # run <command_line_tool>, capture output as it happens
+        self.log(console, 'running <command_line_tool>:')
+        self.log(console, '    '+command_line_tool_cmd_str))
+        p = subprocess.Popen(command_line_tool_cmd_str,
+                    cwd = self.scratch,
+                    stdout = subprocess.PIPE, 
+                    stderr = subprocess.STDOUT, shell = False)
+
+        while True:
```
MrCreosote: Ugh, same.

doc/kb_sdk_impl_methods.md
```diff
@@ -0,0 +1,201 @@
...
+            line = p.stdout.readline()
+            if not line: break
+            self.log(console, line.replace('\n', ''))
+
+        p.stdout.close()
+        p.wait()
```
MrCreosote: Same and same

doc/kb_sdk_impl_methods.md
```diff
@@ -0,0 +1,201 @@
...
+        if p.returncode != 0:
+            raise ValueError('Error running <command_line_tool>, return code: '+str(p.returncode) + 
+                '\n\n'+ '\n'.join(console))
+ ```
+
+[\[Back to Edit Impl\]](#impl)
+
+##### <A NAME="impl-adding-data"></A>5.7. Adding Data To Your Method
+
+Data that is supported by [KBase Data Types](doc/kb_sdk_data_types_table.md) should be added as a workspace object.  Other data that is used to configure a method may be added to the repo with the code.  Large data sets that exceed a reasonable limit (> 1 GB) should be added to a shared mount point.  This can be accomplished by contacting kbase administrators at http://kbase.us.
```
MrCreosote: I'm not sure what this means - does it mean that if you need some data to run your method every time you should stick it in the workspace? If it does, generally speaking, no you shouldn't. If it's small, put it in the repo. If it's large use reference data.

doc/kb_sdk_make_ui.md
```diff
@@ -0,0 +1,148 @@
...
+6. **Specify User Interface**
+7. [Locally Test Module and Method(s)](kb_sdk_local_test_module.md)
+8. [Register Module](kb_sdk_register_module.md)
+9. [Test in KBase](kb_sdk_test_in_kbase.md)
+10. [Complete Module Info](kb_sdk_complete_module_info.md)
+11. [Deploy](kb_sdk_deploy.md)
+
+### <A NAME="ui-widget"></A>6. Specify User Interface
+
+Control of Narrative interaction is accomplished in files in the ui/narrative/methods/<MyMethod> directory.
```
MrCreosote: These are the narrative method specs mentioned earlier.

doc/kb_sdk_make_ui.md
```diff
@@ -0,0 +1,148 @@
...
+
+#### A. Configure the input interface.
+
+The input options are specified in the "parameters" section of the spec.json file. In the following example, the user
+supplies two required parameters, an input name and an output name. By specifying a 'valid_ws_types' the user will
+be present a searchable dropdown of objects that match the specified type. By passing "is_output_name", the user is
+warned if a name with overwrite an existing object or if the name contains invalid characters.
```
MrCreosote: with -> will

doc/kb_sdk_make_ui.md
```diff
@@ -0,0 +1,148 @@
...
+                        "display": "Do not prune",
+                        "id": "none",
+                        "ui_name": "Do not prune"
+                    }
+                ]
+            }
+        },
+```
+
+There are many additional interface options available. One of the best ways to learn is to look at [this gallery](https://narrative.kbase.us/narrative/ws.23109.obj.1) which contains the a variety of KBase apps along with the spec file that generated the user interface.
```
MrCreosote: extraneous the

doc/kb_sdk_make_ui.md
```diff
@@ -0,0 +1,148 @@
...
+In the 'behavior' section of the spec.json, the output of the user interface is mapped to input to your function.
+If you have maintained a consistent naming though these mappings are pretty pro forma but you should make sure
+that you accept and return "workspace" as a narrative_system_variable as shown below.
+```
+	"behavior": {
+		"service-mapping": {
+			"url": "",
+			"name": "MegaHit",
+			"method": "run_megahit",
+			"input_mapping": [
```
MrCreosote: This is not the recommended way to pass object location info to an app - pass one or more references instead. @briehl can show you how to set up the method spec so that references are passed rather than names.

doc/kb_sdk_register_module.md
JamesJeffryes: In other places, I've merely said "Contact Us" and given a link to the Jira board. Would it be better to request developers fill out the linked form directly?

doc/kb_sdk_impl_methods.md
```diff
@@ -0,0 +1,201 @@
...
+10. [Complete Module Info](kb_sdk_complete_module_info.md)
+11. [Deploy](kb_sdk_deploy.md)
+
+#### <A NAME="impl"></A>5. Edit Impl file
+
+In the lib/\<MyModule\>/ directory, edit the <MyModule>Impl.py (or *.pl) "Implementation" file that defines the methods 
+available in the module.  You can follow this guide for interacting with [KBase Data Types](doc/kb_sdk_data_types.md).  
+Basically, the process consists of obtaining data objects from the KBase workspace, and either operating on them 
+directly in code or writing them to scratch files that the tool you are wrapping will operate on.  Result data should 
+be collected into KBase data objects and stored back in the workspace.
```
JamesJeffryes: Sounds good, I was under the impression that DataFileUtils more or less replaced Workspace but the automatic example still uses workspace.

I'll do my best to hunt down the file type specific modules and list them here

doc/kb_sdk_impl_methods.md
```diff
@@ -0,0 +1,201 @@
...
+Please see:
+
+    https://github.com/kbase/kb_sdk/blob/master/doc/kb_sdk_data_types.md
+
+[\[Back to Edit Impl\]](#impl)
+
+
+##### <A NAME="impl-logging"></A>5.3. Logging
+
+Logging where you are is key to tracking progress and debugging.  Our recommended style is to log to a "console" list.  Here is some example code for accomplishing this.
```
JamesJeffryes: Glad to hear I was right to ignore this advice when I was learning.

doc/kb_sdk_impl_methods.md
MrCreosote: I can help here if you need it. BTW DFU doesn't 'replace' the workspace - it wraps it and makes things a bit more simple (like automatic provenance handling). 

Where are you seeing direct WS access in the example? https://github.com/kbase/kb_sdk/blob/master/src/java/us/kbase/templates/module_python_impl.vm.properties

doc/kb_sdk_impl_methods.md
```diff
@@ -0,0 +1,201 @@
...
+        for c in range(bins):
+            report += '   '+str(counts[c]) + '\t--\t' + str(edges[c]) + ' to ' + str(edges[c+1]) + ' bp\n'
+
+        reportObj = {
+            'objects_created':[{'ref':params['workspace_name']+'/'+params['output_contigset_name'], 'description':'Assembled contigs'}],
+            'text_message':report
+        }
+
+        reportName = 'megahit_report_'+str(uuid.uuid4())   # uuid4() instead of getnode() to ensure uniqueness
+        report_obj_info = ws.save_objects({
```
JamesJeffryes: Wait, KBaseReports has a way of saving items directly to the workspace? One of these sections should probably just be a detailed explanation of KBaseReports because I clearly never came to understand it's full potential.

doc/kb_sdk_impl_methods.md
JamesJeffryes: I'm going to need more of introduction to reference data eventually but do you have an example module that uses it?

doc/kb_sdk_impl_methods.md
MrCreosote: That's pretty much the entire point of KBaseReports - you give it the info you want in a report and it bundles it up and saves it into the data stores properly. And yeah, KBaseReports has become important enough to warrant a few paragraphs. Example usage:
https://github.com/kbase/kb_sdk/blob/develop/src/java/us/kbase/templates/module_python_impl.vm.properties#L113-L119

MrCreosote: More complicated usage: https://github.com/kbaseapps/kb_quast/blob/master/lib/kb_quast/kb_quastImpl.py#L179-L191

doc/kb_sdk_impl_methods.md
MrCreosote: Unfortunately I know very little about the reference data system. @rsutormin is the expert, and @scanon knows a lot as well.

doc/kb_sdk_impl_methods.md
JamesJeffryes: So my understanding was that the assemble object is actually saved here: https://github.com/kbase/kb_sdk/blob/develop/src/java/us/kbase/templates/module_python_impl.vm.properties#L107-L110 but that the report simply links to the saved object. Is that incorrect?

MrCreosote: That's correct. The point I'm making is that this and the next few lines of code directly save a report object to the workspace. Instead, the code should use KBaseReport to save the report. The python example module does this correctly.

MrCreosote: Maybe what you're missing here is that a report is a workspace object just like any other workspace object (including the assembly object)? It's just a different type of object. AssemblyUtils handles saving and getting Assembly objects, and KBaseReport handles saving Report objects.

JamesJeffryes: Got it, I had misunderstood "report objects" your initial comment to mean objects in the report not the report itself. Sorry for the digression.

doc/kb_sdk_impl_methods.md
```diff
@@ -0,0 +1,201 @@
...
+6. [Specify User Interface](kb_sdk_make_ui.md)
+7. [Locally Test Module and Method(s)](kb_sdk_local_test_module.md)
+8. [Register Module](kb_sdk_register_module.md)
+9. [Test in KBase](kb_sdk_test_in_kbase.md)
+10. [Complete Module Info](kb_sdk_complete_module_info.md)
+11. [Deploy](kb_sdk_deploy.md)
+
+#### <A NAME="impl"></A>5. Edit Impl file
+
+In the lib/\<MyModule\>/ directory, edit the <MyModule>Impl.py (or *.pl) "Implementation" file that defines the methods
```
JamesJeffryes: Based on the examples, it looks like for java the implementation happens in the server file. I'll note that.

doc/kb_sdk_edit_module.md
```diff
@@ -112,32 +109,37 @@ The type definition can either be another previously defined type name, a primit
...
 
-Optionally, as we have shown in the example, your method can require authentication by adding that declaration at the 
-end of the method.  All methods that run in the Narrative will require authentication because they need to interact with
-a user's workpace).
+All methods that run in the Narrative will require authentication because they need to interact with a user's workspace).
+Your method can require authentication by adding that declaration at the end of the method.
 
-If you will be loading or saving any data from the workspace, make sure you pass in the *workspace_name* in the input 
-parameters.
+If you will be loading or saving any data from the workspace, make sure you accept a workspace reference in the input
```
JamesJeffryes: So it turns out that KBaseReports only accepts workspace_name as a parameter which means that passing only a workspace id causes extra work. I'll open a ticket on this one

JamesJeffryes:
@briehl Would you take a look at the narrative related sections (sdk_make_ui mostly)?
@MrCreosote I've incorporated you recent feedback
@sychan Let me know if you would like the testing, deployment or other sections changed

doc/kb_sdk_make_ui.md
```diff
@@ -0,0 +1,169 @@
...
+6. **Specify User Interface**
+7. [Locally Test Module and Method(s)](kb_sdk_local_test_module.md)
+8. [Register Module](kb_sdk_register_module.md)
+9. [Test in KBase](kb_sdk_test_in_kbase.md)
+10. [Complete Module Info](kb_sdk_complete_module_info.md)
+11. [Deploy](kb_sdk_deploy.md)
+
+### <A NAME="ui-widget"></A>6. Specify User Interface
+
+Control of Narrative interaction is accomplished in files in the ui/narrative/methods/<MyMethod> directory.
```
briehl: The angle brackets will get treated like HTML by markdown and not rendered. so it'll look like `/ui/narrative/methods/` instead. Try using `&lt;` and `&gt;` to escape them.

doc/kb_sdk_make_ui.md
```diff
@@ -0,0 +1,169 @@
...
+
+#### A. Configure the input interface.
+
+The input options are specified in the "parameters" section of the spec.json file. In the following example, the user
+supplies two required parameters, an input name and an output name. By specifying a 'valid_ws_types' the user will
+be present a searchable dropdown of objects that match the specified type. By passing "is_output_name", the user is
```
briehl: typo: should be "the user will be presented with a searchable..."

doc/kb_sdk_make_ui.md
```diff
@@ -0,0 +1,169 @@
...
+warned if a name will overwrite an existing object or if the name contains invalid characters.
```
briehl: Also, the block with that output object will appear at the bottom of the app cell.

doc/kb_sdk_make_ui.md
```diff
@@ -0,0 +1,169 @@
...
+		    "field_type" : "text",
+		    "text_options" : {
+		     	"valid_ws_types" : [ "KBaseGenomes.ContigSet" ],
+		    	"is_output_name":true
+		    }
+		}
+	],
+```
+Another common input is a dropdown which is demonstrated below. For each option, the "value" is what will be passed to
+the script while the UI Name and Display is what the user will see. In this example, the parameter is hidden be default
```
briehl: "passed to the script" should probably be "passed to the App"

Also in this line - the parameter is hidden BY default

doc/kb_sdk_make_ui.md
```diff
@@ -0,0 +1,169 @@
...
+            "allow_multiple": false,
+            "default_values": [ "biochemistry" ],
+            "field_type": "dropdown",
+            "dropdown_options": {
+                "options": [
+                    {
+                        "value": "biochemistry",
+                        "display": "Known Biochemistry",
+                        "id": "biochemistry",
+                        "ui_name": "Known Biochemistry"
```
briehl: I'm pretty sure that only `ui_name` is used at this point, and `display` is redundant. I'll have to test a few things to check, though.

doc/kb_sdk_make_ui.md
```diff
@@ -0,0 +1,169 @@
...
+                        "value": "none",
+                        "display": "Do not prune",
+                        "id": "none",
+                        "ui_name": "Do not prune"
+                    }
+                ]
+            }
+        },
+```
+
```
briehl: It would be nice to have a screenshot of the app cell that this spec produces somewhere around here.

doc/kb_sdk_make_ui.md
```diff
@@ -0,0 +1,169 @@
...
+There are many additional interface options available. One of the best ways to discover them is to look at 
+[this gallery](https://narrative.kbase.us/narrative/ws.23109.obj.1) which contains a variety of KBase apps along with
```
briehl: 👍

doc/kb_sdk_make_ui.md
```diff
@@ -0,0 +1,169 @@
...
+                },
+                {
+                    "narrative_system_variable": "workspace",
+                    "target_property": "workspace_name"
+                }
+            ]
+        }
+    }
+```
+
```
briehl: An example of the transformations might be useful here, too. So, the cell inputs look like one thing, the structure passed to the app looks like another. Likewise with the results - the actual function the app runs returns one specific structure, and that output mapping is used to turn it into another that the viewers can use. Having a concrete example here of what that transform looks like would be nice.

doc/kb_sdk_local_test_module.md
```diff
@@ -74,16 +79,24 @@ Edit the local test config file (`test_local/test.cfg`) with a developer token (
     cd test_local
     kb-sdk test
 
-This will build your Docker container, run the method implementation running in the Docker container that fetches example ContigSet data from the KBase CI database and generates output.  
+This will build your Docker container, run the method implementation running in the Docker 
+container that fetches example ContigSet data from the KBase CI database and generates output.
```
MrCreosote: This sentence seems like it needs some cleanup

doc/kb_sdk_local_test_module.md
```diff
@@ -74,16 +79,24 @@ Edit the local test config file (`test_local/test.cfg`) with a developer token (
     cd test_local
     kb-sdk test
 
-This will build your Docker container, run the method implementation running in the Docker container that fetches example ContigSet data from the KBase CI database and generates output.  
+This will build your Docker container, run the method implementation running in the Docker 
+container that fetches example ContigSet data from the KBase CI database and generates output. 
+After tests are run for the first time, a new `workdir` directory will appear in `test_local`
+that contains the scratch directory which is helpful for inspecting intermediate files,
```
MrCreosote: probably want to mention the scratch dir is in workdir/tmp

doc/kb_sdk_local_test_module.md
```diff
@@ -74,16 +79,24 @@ Edit the local test config file (`test_local/test.cfg`) with a developer token (
...
+Additionally you can inspect the Docker container to see whether your wrapped tool was 
+installed properly, by opening a bash console.  From the `test_local` directory:
     
     ./run_bash.sh
 
-Unfortunately, you will have to rebuild the Docker image each time you change your module code (e.g. KIDL \<MyModule\>.spec, \<MyModule\>Impl.py, and your testing code) but this happens automatically for you when you run *kb-sdk test*, so it just slows you down rather than add any extra effort.  However, if you change the KIDL \<MyModule\>.spec file, you will have to rerun *make* to propagate those changes to the \<MyModule\>Client and \<MyModule\>Impl code (and likely will have some tweaks to apply to the Impl code to match the spec changes).  Happy debugging!
+Unfortunately, you will have to rebuild the Docker image each time you change your module code 
+(e.g. KIDL \<MyModule\>.spec, \<MyModule\>Impl.py, and your testing code) but this happens 
+automatically for you when you run *kb-sdk test*, so it just slows you down rather than add
```
MrCreosote: add -> adds

doc/kb_sdk_make_ui.md
```diff
@@ -0,0 +1,215 @@
...
+### <A NAME="ui-widget"></A>6. Specify User Interface
+
+Control of Narrative interaction is accomplished in files in the ui/narrative/methods/&lt;MyMethod&gt; directory.
+
+#### A. Configure the input interface.
+
+The input options are specified in the "parameters" section of the spec.json file. In the following example, the user
+supplies two required parameters, an input name and an output name. By specifying a 'valid_ws_types' the user will
+be presented with a searchable dropdown of objects that match the specified type. By passing "is_output_name", the user is
+warned if a name will overwrite an existing object or if the name contains invalid characters and a widget displaying that
```
MrCreosote: Should swap the two cases to match up correctly with the end phrase

JamesJeffryes:
@sychan I've addressed the changes @briehl & @MrCreosote requested. Would you mind looking over the docs and merging if you think appropriate?

PR #267, opened 2017-08-11
Travis update

sychan:
Implements a working travis.yaml config in the templated "kb-sdk init" command, as well as putting a tox.ini config in the base config.


.travis.yml
```diff
@@ -1,13 +1,46 @@
-language: java
+language: python
```
MrCreosote: Should be java

src/java/us/kbase/mobu/initializer/ModuleInitializer.java
```diff
@@ -115,6 +115,7 @@ public void initialize(boolean example) throws Exception {
 		Map<String, Path> templateFiles = new HashMap<String, Path>();
 		templateFiles.put("module_typespec", Paths.get(moduleDir, specFile));
 		templateFiles.put("module_travis", Paths.get(moduleDir, ".travis.yml"));
+		templateFiles.put("module_tox", Paths.get(moduleDir, "tox.ini"));
```
MrCreosote: is this needed for java & perl? I'm not sure if it gets populated for those languages because it's in this hash, but it shouldn't.

src/java/us/kbase/templates/module_travis.vm.properties
```diff
@@ -1,6 +1,32 @@
...
+  - pushd ..
+  - git clone https://github.com/kbase/jars
+  - git clone https://github.com/kbase/kb_sdk
+  - cd kb_sdk
+  - make bin
+  - make sdkbase
+  - export PATH=$(pwd)/bin:$PATH
+  - source src/sh/sdk-completion.sh
+  - popd
+  - kb-sdk test -s || true
```
MrCreosote: Probably won't work without checked in tokens

src/java/us/kbase/mobu/initializer/ModuleInitializer.java
sychan: I just checked and tox.ini is definitely python only, I've moved it.

.travis.yml
sychan: Yes, this is leftover cruft

src/java/us/kbase/templates/module_travis.vm.properties
sychan: Getting rid of it.

MrCreosote:
This should go to develop, not master.

PR #273, opened 2017-11-09
Update DFU README.md url to not use my fork

realmarcin:


scanon:
I'm not sure it makes sense to direct them to an empty README for DFU.  It looks like the DFU was already dropped from the list of links.  I'm going to just close this.

scanon:
Wait.  Just saw that you have a PR to DFU that adds docs.  Re-opening.

PR #274, opened 2017-11-10
minor documentation updates

nixonpjoshua:
these are just some small fixes I forgot to make a PR for I made when I was going to through the SDK documentation


doc/kb_sdk_impl_methods.md
```diff
@@ -72,8 +72,8 @@ read from the scratch folder.
 
 ```python
 import os
-from AssemblyUtil.AssemblyUtilClient import AssemblyUtil
-from KBaseReport.KBaseReportClient import KBaseReport
+from Assemblyutil.AssemblyutilClient import AssemblyUtil
```
scanon: This isn't how it show up for me.  Are you sure?

scanon: This doesn't match what I see.  Are you sure?

scanon: Same for the next too.

nlharris: @nixonpjoshua can you reply to @scanon's questions?

nixonpjoshua:
I think that the issues I had were related to the sdk not caring the case of letters I put in for module names while the filesystem did care once files were created for the modules. This caused some errors for me and I had initially thought that the module names were cased differently as by doing that I was able to make my app work. 

I added a warning into the documentation about this pitfall.

alternatively this could be considered a bug in the SDK and the SDK could be made to only match module names in a case-sensitive way which I believe would make this pitfall no longer exist.

nlharris:
I think @thebenallen made some changes after @nixonpjoshua made a few edits. Can you guys check if I merged the changes correctly?

PR #276, opened 2017-11-29
minor edits to text

nlharris:
BTW, I was confused by the wording of "B. Configure passing variables from Narrative Input to SDK method." -- can you perhaps reword that? I'm not sure what it means.


JamesJeffryes (approved):
Looks good

PR #277, opened 2017-11-29
minor improvements to text

nlharris:
I found this hard to understand, but am not sure how to reword it: "Docker will rebuild everything from the first detected change in a dockerfile but pull everything upstream of that from cache. This means that if you are pulling in external data using RUN and a command like `git clone` or `wget` changes in those sources will not automatically be reflected in a rebuilt Docker image unless the Docker file changes at or before that import."


doc/kb_sdk_local_test_module.md
```diff
@@ -33,11 +33,11 @@ RUN \
   make
 ```
 
-Note: you do not have to add any lines to the Dockerfile for installing your SDK Module code, as this will happen automatically.  The contents of your SDK Module git repo will be added at /kb/module.
+Note: you do not have to add any lines to the Dockerfile for installing your SDK Module code, as this will happen automatically.  he contents of your SDK Module git repo will be added at /kb/module.
```
JamesJeffryes: The

JamesJeffryes (changes requested):

PR #279, opened 2017-11-29
minor changes in wording

nlharris:
i'm not sure we still have an "SDK Register" app -- is that still how you register an app? @JamesJeffryes ?


JamesJeffryes:
No you register though the catalog service here https://narrative.kbase.us/#catalog/modules

nlharris:
ok, so we need to change that instruction, then.

nlharris:
(i changed it)

JamesJeffryes (approved):

PR #292, opened 2018-01-24
SCT-475 fix python build, tests, etc.

MrCreosote:
A brand new python user's build currently fails-  there's an incompatibility between sdkbase and the generated Dockerfile.

Current users will likely not see this problem as sdkbase and their module are built already.

* Bump version
* Move all the python SSL warning fixes to sdkbase
* Note that Java 9 is currently incompatible
* Deprecate R.
  * Remove failing R test
* Fix java tests by updating certs
* Minor cleanup

```
$ kb-sdk init -u gaprice -e -l R Foo
************************************************************************
WARNING: R support is deprecated and will be removed in a future release
************************************************************************
Done! Your module is available in the Foo directory.
Compile and run the example methods with the following inputs:
  cd Foo
  make          (required after making changes to Foo.spec)
  kb-sdk test   (will require setting test user account credentials in
test_local/test.cfg)

  
$ cd Foo
$ make
kb-sdk compile Foo.spec \
		--out lib \
		--plclname Foo::FooClient \
		--jsclname javascript/Client \
		--pyclname Foo.FooClient \
		--rclname Foo/FooClient \
		--javasrc src \
		--java \
		--rsrvname Foo/FooServer \
		--rimplname Foo/FooImpl;
KBase SDK version 1.0.18 (commit 389297a)
************************************************************************
WARNING: R support is deprecated and will be removed in a future release
************************************************************************
  
$ gedit test_local/test.cfg 
$ kb-sdk test
************************************************************************
WARNING: R support is deprecated and will be removed in a future release
************************************************************************

Validating module in (/home/crusherofheads/temp/Foo)
```

### DIFF
--- a/DeveloperNotes.md
+++ b/DeveloperNotes.md
@@ -1,0 +1,3 @@
+Regarding Java 9 incompatibility: https://blog.codefx.org/java/jsr-305-java-9/
+
+Specifically, the Generated annotation is a problem.
--- a/README.md
+++ b/README.md
@@ -9,7 +9,7 @@ There are still some general restrictions on functionality that will gradually b
 - Operates only on supported KBase data types
 - Uses existing data visualization widgets
 - Does not require new uploaders/downloaders
-- Wrapper written in Python, Java, R, or Perl
+- Wrapper written in Python, Java, or Perl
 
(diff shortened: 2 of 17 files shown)


MrCreosote:
Note that we'll need to update sdkbase in the various environments

PR #293, opened 2018-01-26
Other fixes from Gavin

scanon:
I merged before seeing Gavin's earlier comments.  This addresses those.


MrCreosote:
Looks good to me

PR #295, opened 2018-01-30
added some info about useful sdk modules

nlharris:
from gavin's cheatsheet, "SDK Modules that are probably useful for your SDK module development"


MrCreosote:
Point the first - this should go against the develop branch

Point the second - should be integrated into this overlapping list: https://github.com/kbase/kb_sdk/blob/develop/doc/kb_sdk_impl_methods.md#a-install-other-kbase-modules

PR #297, opened 2018-01-31
Update kb_sdk_dockerized_install.md

bio-boris:
Fix docker installation instructions.


MrCreosote:
Looks like there's some weirdness re the ref data stuff, but that should've gone into develop in the first place, so I think it's ok.

PR #298, opened 2018-01-31
minor changes

nlharris:
there may be more changes coming when i merge in bits of the sdk "cheat sheet" but i don't want to lose these.


MrCreosote:
This needs to go against develop

PR #301, opened 2018-02-05
integrated section "Call one SDK module from another" from cheatsheet

nlharris:


doc/kb_sdk_impl_methods.md
```diff
@@ -87,6 +87,38 @@ class <ModuleName>:
         #END_CONSTRUCTOR
         pass
 ```
+##### Calling one SDK module from another
```
MrCreosote: This is already covered in this section as far as I can tell and is mostly redundant. If there are improvements that need to be made they should be integrated into the existing text.

nlharris:
This is a summary that pulls together into one place various bits of info that is scattered around. I think it's still useful as a summary, but we can leave it out if you prefer.

MrCreosote:
all the info is on this page isn't it? Are there missing parts?

MrCreosote:
BTW I'm not saying this page doesn't need improvement.

doc/kb_sdk_impl_methods.md
scanon: It seems like there is some additional information here but maybe it should be trimmed.  For example, we should show how to find documentation on the modules.  So I think we just need to add some text from step 1 below and maybe step 7.  Also, if we add it, it should reference narrative.kbase.us instead of ci.kbase.us.

jayrbolton:
:+1:

PR #302, opened 2018-02-08
Update kb_sdk_impl_methods.md

bio-boris:
Make it more clear to use more files for your code otherwise write your functions in between the magic comments

### DIFF
--- a/doc/kb_sdk_impl_methods.md
+++ b/doc/kb_sdk_impl_methods.md
@@ -15,9 +15,7 @@
 ### <A NAME="impl"></A>5. Implement App(s)
 
 In the directory lib/\<MyModule>, edit the \<MyModule>Impl.py (or *Impl.pl or *Server.java) "Implementation" file that 
-defines the methods (apps) available in the module. The example module is very simple and is implemented directly in this file
-but it is better practice and more readable to separate implementation logic into multiple files, especially for modules 
-that include more than one app. 
+defines the methods (apps) available in the module. It is important to notice that only the code that is written **in between the comments** in the implementation file will be kept, and the rest will be discarded. The example module is very simple, so its logic is located exclusively in one file; However it is better practice and more readable to separate implementation logic into multiple files, especially for modules that include more than one app.
 
 The workflow of most modules involves obtaining data from the KBase data stores, operating on that data (possibly with 3rd-party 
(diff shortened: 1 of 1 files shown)


doc/kb_sdk_impl_methods.md
```diff
@@ -15,9 +15,7 @@
 ### <A NAME="impl"></A>5. Implement App(s)
 
 In the directory lib/\<MyModule>, edit the \<MyModule>Impl.py (or *Impl.pl or *Server.java) "Implementation" file that 
-defines the methods (apps) available in the module. The example module is very simple and is implemented directly in this file
-but it is better practice and more readable to separate implementation logic into multiple files, especially for modules 
-that include more than one app. 
+defines the methods (apps) available in the module. It is important to notice that only the code that is written **in between the comments** in the implementation file will be kept, and the rest will be discarded. The example module is very simple, so its logic is located exclusively in one file; However it is better practice and more readable to separate implementation logic into multiple files, especially for modules that include more than one app.
```
MrCreosote: I feel like being more specific about what the comments are or look like would help newbies.

bio-boris: @MrCreosote I agree. 

My shot at it would be like:

Magic comments are comments that are used internally by the kbase_sdk in order to do XXXX.

Examples of magic comments include

#BEGIN_HEADER
(This is where your import statements go)
#END_HEADER

#BEGIN_CLASS_HEADER
(This is where your class variables and functions go that you want imported)
#END_CLASS_HEADER

#BEGIN_CONSTRUCTOR
(This is in your _init_ statement for your class goes)
#END_CONSTRUCTOR

 #BEGIN YourFunctionName1
(This is were the implementation details of your functions go)
#END YourFunctionName1

 #BEGIN YourFunctionName2n
(This is were the implementation details of your functions go)
#END YourFunctionName2n

Any code created outside of the Magic Comments will not be included inside the final .impl implementation file.

MrCreosote: Looks good. A few notes:

1) Perl doesn't have the class header section
2) `(This is where your class variables and functions go that you want imported)`
Is imported the right word here?

The way I would put the last sentence is something like

When the implementation file is recompiled, any code outside the magic comments will be lost, and recompilation occurs on any build or test.

PR #304, opened 2018-02-15
Change name of the base image that is built/used

sychan:


sychan:
I ran the following all without error:
make
make sdkbase
 (with the new kb-sdk)
kb-sdk init --example -l python -u sychan MyContigs5
kb-sdk test (against CI)

entrypoint
```diff
@@ -21,7 +21,7 @@ EOF
 elif [ "z$1" = "zsdkbase" ] ; then
     echo "Pulling and tagging the base image"
     docker pull kbase/sdkbase
```
scanon: Shouldn't this pull kbase/kbase:sdkbase2.latest?

sychan: I don't really know the distinction between kbase/kbase.sdkbase.latest and kbase/sdkbase

Are they the same image, but under different tags?

sychan:
@scanon I'm guessing that we pushed kbase/kbase:sdkbase.latest to dockerhub as kbase/sdkbase:latest? I've gotten rid of the phantom image renaming and just push it to dockerhub as kbase/kbase:sdkbase.latest and pull it back down again.

entrypoint
```diff
@@ -20,8 +20,7 @@ exec docker run -it --rm -v \$HOME:\$HOME -u \$(id -u) -w \$(pwd) -v /run/docker
 EOF
 elif [ "z$1" = "zsdkbase" ] ; then
     echo "Pulling and tagging the base image"
-    docker pull kbase/sdkbase
-    docker tag kbase/sdkbase kbase/kbase:sdkbase.latest
+    docker pull kbase/kbase:sdkbase2.latest
```
MrCreosote: is the tag line supposed to be deleted?

Why does the image name change other than appending a 2?

MrCreosote:
```
~/localgit/kb_sdk$ make test
git submodule init
git submodule update

*snip*

     [exec] 0	us.kbase.tools.WorkspaceTools.GlobalArgs (covered=0/3)
     [exec] 0	us.kbase.tools.WorkspaceTools.HelpCommandArgs (covered=0/6)
     [exec] Common coverage: 72.4878(38217/52722)

BUILD SUCCESSFUL
Total time: 18 minutes 16 seconds
~/localgit/kb_sdk$ git log
commit e29f39ed467111aa60bdca74b3d45439d76e71c2
Author:  *snip*
Date:   Thu Feb 15 15:57:35 2018 -0800

    Modify entrypoint to pull from new image name

entrypoint
sychan: Because the old image name seems to only have that name due to it being retagged and then pushed up. Since I just pushed kbase/kbase:sdkbase.latest without the rename, there isn't a reason to revert it to the old name.

There is kbase/sdkbase:develop up there now, but that is the "built from scratch" image, not the patched for python image built by kb_sdk/sdkbase/Dockerfile

MrCreosote:
This looks ok to me, but I'm not super knowledgeable about this chunk of the system. @scanon if you're happy go ahead and merge.

scanon:
I think we should merge it.

PR #309, opened 2018-04-13
Docs update

jayrbolton:
This is a large docs-only update that does the following:
* Turns the SDK guide into more of a detailed step-by-step tutorial where you build a real app -- i.e. it guides you through each little step in creating the ContigFilter example, rather than just using it as a reference
* Adds an additional step at the beginning of the tutorial that goes over some basic sdk concepts
* Adds a How-to section with a starting collection of articles, most of which were pulled from later reference-oriented sections in the SDK guide

I didn't remove any content from the old guide, but some of it is edited or moved into the How-to sections.

All tutorial content lives under `doc/tutorial` and all how-to content lives under `doc/howto`


doc/howto/edit_your_dockerfile.md
```diff
@@ -0,0 +1,49 @@
+# How to edit your app's Dockerfile
+
+To set up custom programs for your app, such as third-party tools, you can edit your Dockerfile, which contains the setup code needed to build up your container.
+
+A helpful resource is the [Dockerfile Best Practices](https://docs.docker.com/engine/userguide/eng-image/dockerfile_best-practices) guide.
+
+The base KBase Docker image contains a KBase Ubuntu image, but not much else. You will need to include any dependencies, including the installation of whatever tool you are wrapping, to build a custom Docker image that can run your Module.
```
sychan: Actually the KBase Docker image also contains the dependencies necessary to run a JSON-RPC server in the supported language, as well as a core set of KBase API clients

doc/howto/edit_your_dockerfile.md
```diff
@@ -0,0 +1,49 @@
...
+```
+RUN wget blah.com/tarball.tgz
+RUN tar xzf tarball.tgz
+RUN cd tarball
+RUN make
+RUN cd ..
+RUN rm -rf tarball.tgz
+```
+
+because the latter does not remove the previous bloating layers, despite the "rm" command.
```
sychan: Might be better to call out that each "RUN" command adds an additional layer, which is why so many shell commands are chained together in a single RUN statement. Also point out that the && operator means if any commands returns a exit code that isn't strictly a success code, the chain of commands will stop at that point.

jayrbolton:
Made a couple changes
- updated the Dockerfile Howto article 
- I noticed I accidentally left old content in this step of the tutorial that I refactored: https://github.com/jayrbolton/kb_sdk/blob/jayrbolton-docs/doc/tutorial/impl_methods.md
I think that step will be less confusing now

PR #313, opened 2018-05-07
Virtualize tests

jayrbolton:
This is a start at getting the tests to run in an Ubuntu VM using Vagrant. The Vagrant image can be found here: https://app.vagrantup.com/kbase/boxes/kb-sdk-test. The tests can be run with `make test-vagrant`.

I updated the `test_dependencies.md` document to add a lot more details and organization.

I haven't yet gotten the tests to work and currently get a variety of failures. I think the only way to work towards fixing them would be to "wolf fence" out all the tests and get them to run one section at a time -- lots of work.


jayrbolton:
Updated so base is develop instead of master

jayrbolton:
Poked at this some more and got some more tests passing. Still some failing tests with Perl, callback server, etc. Test logs here: https://gist.github.com/jayrbolton/32a7423c61e52ae8cadfaf6058c3cbaf

Right now the output for these tests is really verbose, almost 9,000 lines long, and it's very hard to see what's happening. Would like to reduce this

jayrbolton:
@sychan maybe we can merge this? Would like to build on this virtualbox image provided here in other branches (eg the python3 branch). Also, maybe we should merge `master` into `develop` to get the docs updates that I made, which went into master directly. I'd like to do a docs update that links the new sphinx website.

test_dependencies.md
```diff
@@ -1,21 +1,75 @@
...
+
+## Python
+
+```sh
+sudo apt-get install python python-setuptools
+easy_install pip
+pip install requests nose2 jsonrpcbase
+```
+
+New versions of paramiko complain about being unable to import SSH2_AGENTC_REQUEST_IDENTITIES, 1.10.2 works.
```
MrCreosote: Looks like the paramiko install has been removed, but this line is still here. Do we need paramiko with the new clients?

test_dependencies.md
```diff
@@ -1,21 +1,75 @@
...
+$ sudo cpan -f -i RPC::Any
+$ sudo cpan Bio::KBase::Exceptions
+$ sudo cpan Parse::Yapp Devel::StackTrace Lingua::EN::Inflect Template File::Slurp Cwd JSON Data::UUID XML::Dumper JSON::RPC::Client Exception::Class Config::Simple Digest::SHA1 Crypt::OpenSSL::RSA Convert::PEM DateTime MIME::Base64 URI Object::Tiny::RW Plack File::ShareDir::Install YAML TAP::Harness Plack::Middleware::CrossOrigin RPC::Any::Server::JSONRPC::PSGI
+```
+
+Must force RPC::Any, Moose updates cause XML tests to fail.
+
+## Misc
+
+With the new auth2 clients and server we may no longer need rsa, paramiko, Crypt::OpenSSL::RSA, Convert::PEM
```
MrCreosote: Not installing rsa/paramiko anymore, should we be?

sychan: If the tests pass then Paramiko may be unneeded at this point. IIRC Paramiko was useful for some ssh-key based authentication features in the old auth2, but I don't think anyone ever used that

MrCreosote: Not sure what the perl clients still use though.

PR #314, opened 2018-05-08
Add `kb-sdk shell` command

jayrbolton:
Here is my first shot at adding some Java. @MrCreosote let me know what feedback you have.

This opens the docker shell as a root user, which I know Shane didn't want. But it works well for me. We can update this PR later with an alternative UID. Also I think we could remove the `run_bash.sh` script once this update looks good.


src/java/us/kbase/mobu/shell_launcher/ShellLauncher.java
```diff
@@ -0,0 +1,48 @@
...
+import java.util.Map;
+import java.nio.file.Paths;
+
+import org.yaml.snakeyaml.Yaml;
+
+import us.kbase.mobu.util.DirUtils;
+import us.kbase.mobu.util.TextUtils;
+import us.kbase.mobu.ModuleBuilder;
+
+public class ShellLauncher {
```
MrCreosote: Needs javadocs

MrCreosote: blank line after class declaration

src/java/us/kbase/mobu/shell_launcher/ShellLauncher.java
```diff
@@ -0,0 +1,48 @@
...
+    File moduleDir;
```
MrCreosote: Class fields should be `private final` unless there's a good reason otherwise

MrCreosote: Use `Path` from nio, File is out of date.

src/java/us/kbase/mobu/shell_launcher/ShellLauncher.java
```diff
@@ -0,0 +1,48 @@
...
+    String moduleName;
+    String workDir;
```
MrCreosote: Use `Path`

src/java/us/kbase/mobu/shell_launcher/ShellLauncher.java
```diff
@@ -0,0 +1,48 @@
...
+        String kbaseYml = TextUtils.readFileText(new File(moduleDir, "kbase.yml"));
+        @SuppressWarnings("unchecked")
+        Map<String,Object> config = (Map<String, Object>)new Yaml().load(kbaseYml);
+        moduleName = (String)config.get("module-name");
+        workDir = Paths.get(moduleDir.toPath().toString(), "test_local", "workdir").toString();
+    }
+
+    public Process exec() throws IOException, InterruptedException {
+        // This command looks like:
+        // `docker run -i -t -v WORKDIR:/kb/module/work test/module:latest bash`
```
MrCreosote: Are there long forms of the flags? Would be easier to read if so. If not, document what the flags mean

src/java/us/kbase/mobu/ModuleBuilder.java
```diff
@@ -29,6 +29,7 @@
 import us.kbase.mobu.util.ProcessHelper;
```
MrCreosote: Is there any way to automatically test this?

src/java/us/kbase/mobu/ModuleBuilder.java
```diff
@@ -96,6 +98,10 @@ public static void main(String[] args) throws Exception {
         RunCommandArgs runArgs = new RunCommandArgs();
         jc.addCommand(RUN_COMMAND, runArgs);
 
+        // add the 'shell' command
+        ShellCommandArgs shellArgs = new ShellCommandArgs();
```
MrCreosote: `final`

src/java/us/kbase/mobu/ModuleBuilder.java
```diff
@@ -437,6 +445,18 @@ private static int runRunCommand(RunCommandArgs runArgs, JCommander jc) {
         }
     }
 
+    private static int runShellCommand(ShellCommandArgs shellArgs, JCommander jc) {
```
MrCreosote: `final` args

src/java/us/kbase/mobu/ModuleBuilder.java
```diff
@@ -437,6 +445,18 @@ private static int runRunCommand(RunCommandArgs runArgs, JCommander jc) {
...
+        try {
+            ShellLauncher launcher = new ShellLauncher();
+            launcher.exec();
+            return 0;
+        } catch (Exception e) {
+            e.printStackTrace();
```
MrCreosote: Add a verbose argument like the other commands, and only print the trace if true

MrCreosote: Alternatively just pass the exception and verbose boolean to showError() and put all the error display logic in one spot

src/java/us/kbase/mobu/ModuleBuilder.java
```diff
@@ -137,7 +143,9 @@ public static void main(String[] args) throws Exception {
 	        returnCode = runInstallCommand(installArgs, jc);
 	    } else if (jc.getParsedCommand().equals(RUN_COMMAND)) {
 	        returnCode = runRunCommand(runArgs, jc);
-	    }
+	    } else if (jc.getParsedCommand().equals(SHELL_COMMAND)) {
```
MrCreosote: There's some inconsistent indenting going on here

src/java/us/kbase/mobu/shell_launcher/ShellLauncher.java
```diff
@@ -0,0 +1,48 @@
...
+public class ShellLauncher {
+    File moduleDir;
+    String moduleName;
+    String workDir;
+
+    public ShellLauncher() throws IOException {
+        moduleDir = DirUtils.findModuleDir();
+        String kbaseYml = TextUtils.readFileText(new File(moduleDir, "kbase.yml"));
+        @SuppressWarnings("unchecked")
+        Map<String,Object> config = (Map<String, Object>)new Yaml().load(kbaseYml);
```
MrCreosote: Do `new Yaml(new SafeConstructor())`

src/java/us/kbase/mobu/shell_launcher/ShellLauncher.java
```diff
@@ -0,0 +1,48 @@
...
+        moduleName = (String)config.get("module-name");
```
MrCreosote: Check for null in case the yaml file is corrupted

src/java/us/kbase/mobu/shell_launcher/ShellLauncher.java
```diff
@@ -0,0 +1,48 @@
...
+import us.kbase.mobu.util.TextUtils;
+import us.kbase.mobu.ModuleBuilder;
+
+public class ShellLauncher {
+    File moduleDir;
+    String moduleName;
+    String workDir;
+
+    public ShellLauncher() throws IOException {
+        moduleDir = DirUtils.findModuleDir();
```
MrCreosote: It seems like all this stuff is something that's going to happen over and over - getting the module dir, module name, and work dir. Maybe make a class that does all that stuff in one place?

src/java/us/kbase/mobu/shell_launcher/ShellLauncher.java
```diff
@@ -0,0 +1,48 @@
...
+        String kbaseYml = TextUtils.readFileText(new File(moduleDir, "kbase.yml"));
+        @SuppressWarnings("unchecked")
+        Map<String,Object> config = (Map<String, Object>)new Yaml().load(kbaseYml);
+        moduleName = (String)config.get("module-name");
+        workDir = Paths.get(moduleDir.toPath().toString(), "test_local", "workdir").toString();
+    }
+
+    public Process exec() throws IOException, InterruptedException {
```
MrCreosote: Returning the process seems really weird to me. I would make the method `void` unless there's a good reason otherwise.

src/java/us/kbase/mobu/ModuleBuilder.java
MrCreosote: Have you tested with mac & linux, and do you see any of the issues described in the `--user` ticket?

src/java/us/kbase/mobu/shell_launcher/ShellLauncher.java
MrCreosote: If the yaml is corrupted or isn't a map this will throw a nasty error

MrCreosote (changes requested):

jayrbolton:
Thanks for the review. I took a lot of the code around getting the module name/directory from elsewhere ([for example here](https://github.com/kbase/kb_sdk/blob/master/src/java/us/kbase/mobu/renamer/ModuleRenamer.java#L45)). It felt pretty overly complicated and seems like it should be a lot easier. @MrCreosote maybe you have better boilerplate for passing/fetching module name/directory from the `ShellLauncher` class?

MrCreosote:
Not off the top of my head. The first thing I'd do is isolate all that stuff in a `ModuleInfo` class (or something) so if there's a way to simplify it can be done in one place rather than all over.

src/java/us/kbase/mobu/ModuleBuilder.java
jayrbolton: Running as root causes any files you create in the container to be owned by "root" in the host. To me this is not a problem and is something I would expect with docker. But Shane said it was a usability issue for people. I'll to talk to Shane about it soon in person.

src/java/us/kbase/mobu/ModuleBuilder.java
jayrbolton: Unfortunately there is a big mix of tabs and spaces throughout ModuleBuilder.java. I changed it to match the "formatting" of the rest of this section (ie. a tab character and then spaces afterward).

eg. line 151 is `<tab-character><four-spaces><tab-character>System.exit(returnCode);`

:man_shrugging:

jayrbolton:
@MrCreosote I made changes and would be ready for a re-review at any point, no hurry. Change notes:
- Added javadocs comments
- Returning void rather than Process in ShellLauncher#exec()
- Converted a lot of my variables to final (I think all of them)
- Added a `ModuleInfo` class that gets module config and directories. Has some better error handling.
- Replaced all tab characters in ModuleBuilder with four spaces

Other notes:
- I will test on a mac soon
- Can do a RELASE_NOTES.txt update for this and increase patch semver. Not sure what conventions you all have here. 
- Lots of other files mix tabs and spaces. I can fix all of them in another PR.
- There is a lot of opportunity to refactor code all over the place here. Will be tricky to balance big refactors vs incremental changes on this repo

MrCreosote:
> I will test on a mac soon
We should do that before merge.

>Can do a RELASE_NOTES.txt update for this and increase patch semver. Not sure what conventions you all have here.

Just follow the conventions in the current file, but make the release date TBD. We'll add that when we merge to master.

> Lots of other files mix tabs and spaces. I can fix all of them in another PR.

Dooo eet

> There is a lot of opportunity to refactor code all over the place here. Will be tricky to balance big refactors vs incremental changes on this repo

Refactor as you go IMO, otherwise you'll be refactoring for the next few months. If you're working on a chunk of code, make it better.

src/java/us/kbase/mobu/ModuleBuilder.java
MrCreosote: What happens if you run the container as root and produce files owned by root on a system where you don't have root access? Seems like you wind up with a bunch of files in your home directory you can't touch.

MrCreosote: Still have the same question here.

jayrbolton:
Still todo on this one:
- [ ] test on mac
- [ ] take some kind of stab at making tests for the classes created here (ModuleInfo, at least)

jayrbolton:
@sychan This could potentially be merged. I've tested it quite a bit, including on mac. The remaining thing I wanted to do was take a stab at writing tests for the `ModuleInfo` class I added here, but didn't make a ton of headway. Trying to write tests for that class might not be worth it right now

jayrbolton:
Diff without the whitespace changes to ModuleBuilder: https://github.com/kbase/kb_sdk/pull/314/files?w=1

jayrbolton:
@scanon maybe you have feedback on this one

jayrbolton:
:man_shrugging:

PR #315, opened 2018-05-08
Python3

JamesJeffryes:
These are my changes to make the kb_sdk generate python3 modules. It replaces the current python2 templates with python3 versions and sets up the module's dockerfile to inherit from a python3 base image which uses miniconda3 as the default python interpreter

I was never able to get all the tests to work correctly. The nose (not-J unit) python tests run and I confirmed functionality by building and testing a SDK module (Filter Contigs) and a dynamic service (Biochemistry API) in Python3

### DIFF
--- a/src/java/us/kbase/templates/module_dockerfile.vm.properties
+++ b/src/java/us/kbase/templates/module_dockerfile.vm.properties
@@ -1,4 +1,8 @@
+#if( $language == "python")
+FROM jjeffryes/kbasepythonsdk:latest
+#else
 FROM kbase/kbase:sdkbase.latest
+#end
 MAINTAINER #if($username)${username}#{else}KBase Developer#{end}
 
 # -----------------------------------------
--- a/.gitignore
+++ b/.gitignore
@@ -10,6 +10,7 @@
 /TestAsyncPerl/
 /reports/
 /test_scripts/ee_mock_service/
+/test_scripts/test.cfg
 /*/*.pyc
 /*/*/*.pyc
 /*/*/*/*.pyc
--- a/lib/biokbase/common/filetools.py
+++ b/lib/biokbase/common/filetools.py
@@ -41,7 +41,7 @@ class ExcelDictReader(object):
                 raise IndexError('Workbook has no sheet for index {}'.format(
                     sheet))
             self._sheet = self._xl.sheet_by_index(sheet)
-        elif isinstance(sheet, basestring):
+        elif isinstance(sheet, str):
             if sheet not in self._xl.sheet_names():
                 raise ValueError('No such sheet name: {}'.format(sheet))
             self._sheet = self._xl.sheet_by_index(sheet)
@@ -60,7 +60,7 @@ class ExcelDictReader(object):
         if not self._sheet:
             raise StopIteration('No sheet set')
         s = self._sheet
-        for i in xrange(1, s.nrows):
+        for i in range(1, s.nrows):
             self._rownum = i + 1
             r = convert_xl_row(self._xl, self._sheet.name, i)
             ret = {n: v for n, v in zip(self._heads, r)}
--- a/lib/biokbase/common/itertools.py
+++ b/lib/biokbase/common/itertools.py
@@ -19,7 +19,7 @@ def chunkiter(iterable, size):
   """
     def inneriter(first, iterator, size):
         yield first
-        for _ in xrange(size - 1):
+        for _ in range(size - 1):
             yield iterator.next()
     it = iter(iterable)
     while True:
--- a/lib/biokbase/log.py
+++ b/lib/biokbase/log.py
@@ -67,7 +67,7 @@ METHODS
 """
 
 import json as _json
-import urllib2 as _urllib2
+import urllib.request as _urllib
 import syslog as _syslog
 import platform as _platform
 import inspect as _inspect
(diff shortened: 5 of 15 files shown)


JamesJeffryes:
@sychan @jayrbolton Here's my Python3 work. Seemed to work for python but never got the tests suite to pass

jayrbolton:
Thanks @JamesJeffryes -- I will work through this today or tomorrow

jayrbolton:
This is looking good to me. I will take a shot at trying to get the tests to pass

jayrbolton:
A couple other conveniences that I can add:
- [ ] dependency tracking (eg. pipenv and Pipfile)
- [ ] flake8 validation and tox.ini

sychan:
@jayrbolton Feel free to merge when you want. There should already be a tox.ini config for flake8 in the kb_sdk develop branch (if not in master already).

jayrbolton:
Pending PR for james on his branch here: https://github.com/JamesJeffryes/kb_sdk/pull/1
adds the things i mentioned above

jayrbolton:
Added another PR for @JamesJeffryes with some additions: https://github.com/JamesJeffryes/kb_sdk/pull/2

Some more things to consider before merging:
* @JamesJeffryes Do you have a repo that saves shell commands on a base ubuntu image to create `kbasepythonsdk`? Might be worth linking / moving somewhere. Also we'll need to move that into the kbase org on docker hub. I could also go through the image and see if I can find any cleanups
* I don't know about the py tests in this repo. Might be worth reviewing them and seeing if they cover the changes we made

jayrbolton:
I think this one can be merged. @MrCreosote Do you want to check this out first?

src/java/us/kbase/templates/module_dockerfile.vm.properties
```diff
@@ -1,4 +1,8 @@
+#if( $language == "python")
+FROM jjeffryes/kbasepythonsdk:latest
```
sychan: We don't want to use a base image coming out of James' personal docker area. I have an image used for Python core services at kbase/kb_python:latest, maybe we should merge James' python3 SDK stuff into that image and just that one. Or else just put the python3 support into sdkbase2?

JamesJeffryes: I think we want to put it just into kbase/kb_python:latest because we want the default python interpreter to be Py3 in this image but we don't want that in sdkbase2. Do you have a branch I can submit a PR against for the kbase/kb_python dockerfile?

JamesJeffryes: This is my branch of sdkbase2 refactored for conda3: https://github.com/JamesJeffryes/sdkbase/tree/python

It looks like your kb_python branch does not have all the support libraries

MrCreosote:
If @sychan @JamesJeffryes and @jayrbolton all agree this is ok, then I won't do a review, but can someone summarize what steps we've taken to ensure that python2 modules are unaffected (or if they are, what needs to happen when someone upgrades their SDK) and that py3 modules work? Do tests now pass, and are there py3 tests?

Also, do we have docs describing how to build a py3 module?

jayrbolton:
Regarding the above questions:
- Test files have been updated (eg test_auth_client.py) to be compatible with python3 and all python tests pass
- This should still work fine with existing python2 apps as long as the app is still using the `sdkbase2` image. If they manually update their app dockerfile to use this python3 image, then their app tests will fail unless everything works on py3
- I can create a PR on kb_sdk_docs to add any information about python3, and we could merge that when this eventually goes live.

I am hitting some bugs using this branch of the sdk when working with both new and existing apps that I didn't see before. Will investigate..

MrCreosote:
> Test files have been updated (eg test_auth_client.py) to be compatible with python3 and all python tests pass

With both py2 and py3? There are py2 and py3 tests right, including the client/server tests?

> I can create a PR on kb_sdk_docs to add any information about python3, and we could merge that when this eventually goes live.

Sounds like a plan

jayrbolton:
I added a test runner that runs all the python tests using both python2 and python3 and all pass (you can run them with `make test-vagrant`). I still haven't looked in detail at the actual tests -- there are only 7 cases.

src/java/us/kbase/templates/module_dockerfile.vm.properties
jayrbolton: Sounds like it might be worth keeping a separate image/repo for kbasepythonsdk. Also was wondering if we really need a full clone of all those repos for SDK apps to run (I mean [these lines](https://github.com/JamesJeffryes/sdkbase/blob/python/Dockerfile#L107?))?

MrCreosote:
> I added a test runner that runs all the python tests

What do you mean by python tests? Tests that are written in python? Or tests that test the python clients & servers/apps?

jayrbolton:
Remaining todos for this PR:
- Need to move the `jjeffryes/kbasepythonsdk`  image into the kbase org on docker hub
- I am not sure about any java tests that might test python clients/servers. All I know so far is that any tests written in python are good to go. 

Regarding tests, I can only say that manual tests I've done here have worked. This includes testing both new and existing apps using the compiled binary from this branch.

src/java/us/kbase/templates/module_dockerfile.vm.properties
sychan: I don't think the whole repo is being kept around - the repos are cloned into /tmp and then the necessary files copied out and then the repos are deleted. I would prefer that we packaged our kbase python modules as pip-installable modules, but we don't have that, so we have these hairballs.

PR #316, opened 2018-05-10
Remove makefile compilation flags to generate client code

jayrbolton:
I've tested things manually only by generating/running/testing some apps -- all looked good.

I left the `--java` flag. It seems to generate java client code in `/src`. It looks like `--javasrc src` def needs to be in there for the java case.

### DIFF
--- a/src/java/us/kbase/templates/module_makefile.vm.properties
+++ b/src/java/us/kbase/templates/module_makefile.vm.properties
@@ -29,14 +29,6 @@ all: compile build build-startup-script build-executable-script build-test-scrip
 compile:
 	kb-sdk compile $(SPEC_FILE) \
 		--out $(LIB_DIR) \
-		--plclname $(SERVICE_CAPS)::$(SERVICE_CAPS)Client \
-		--jsclname javascript/Client \
-		--pyclname $(SERVICE_CAPS).$(SERVICE_CAPS)Client \
-#if($language == "r")
-		--rclname $(SERVICE_CAPS)/$(SERVICE_CAPS)Client \
-#end
-		--javasrc src \
-		--java \
 #if($language == "python")
 		--pysrvname $(SERVICE_CAPS).$(SERVICE_CAPS)Server \
 		--pyimplname $(SERVICE_CAPS).$(SERVICE_CAPS)Impl;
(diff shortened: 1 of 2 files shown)


src/java/us/kbase/templates/module_makefile.vm.properties
```diff
@@ -49,6 +41,8 @@ compile:
 #if($language == "java")
 		--javasrv \
 		--javapackage ${java_package_parent};
+		--javasrc src \
+		--java \
```
MrCreosote: Why generate java clients for java servers when we don't generate py clients for py servers etc?

jayrbolton: I left it in because I was not able to grok the purpose of that flag, but also I am fairly confident we can take it out as long as `--javasrv` is left

So the flag sets javaClientSide in ModuleBuilder and then is passed to here: https://github.com/kbase/kb_sdk/blob/master/src/java/us/kbase/mobu/compiler/RunCompileCommand.java#L151. javaClientSide will always be true with the --javasrv flag. It's unclear to me how any of that is connected to JavaTypeGenerator.processSpec.

jayrbolton:
Item still on my todo-list for this PR: compile and use this to test some existing apps for all languages.

jayrbolton:
Did a little testing on this today with random apps. It appears that if you remove the `--pyclname` flag, it still generates `authclient.py`, `baseclient.py`, and `ModuleClient.py` directly in your `lib/` dir, instead of in your `lib/Module` dir. So that doesn't work. I'll have to do some poking around. It's a bit difficult to trace the call path for each of these flags.

jayrbolton:
Will close this for now so less PRs are open. Python boilerplate code still gets injected regardless of the `pyclname` flag -- need to try to stack trace some stuff :grimacing:

jayrbolton:
Reopening this -- @JamesJeffryes maybe you could add the template generator code changes here?

JamesJeffryes:
@jayrbolton Opened a PR against your branch for this. 

This functionality is tested hit by `src/java/us/kbase/mobu/initializer/test/InitializerTest.java` but it only looks for server files so this test still passes without changes

scanon:
Can we merge this?

jayrbolton:
@scanon Yes I think it's ready. I can work on merging and checking things unless you want to.

scanon:
@jayrbolton You go ahead.

PR #320, opened 2018-07-18
remove -user $(id -u) setting which causes issues running local tests

JamesJeffryes:
This removes the -user setting so the containers run as root. As noted, this might generate files in the test_local dir that the user then does not own but only on Linux. Furthermore, these files are perged on each test run by this code here:
https://github.com/kbase/kb_sdk/blob/2c5a6cd346256b28e4b202d4465665e64857f2d4/src/java/us/kbase/mobu/tester/ModuleTester.java#L165-L171

### DIFF
--- a/src/java/us/kbase/templates/module_makefile.vm.properties
+++ b/src/java/us/kbase/templates/module_makefile.vm.properties
@@ -7,6 +7,7 @@ LIB_DIR = lib
 SCRIPTS_DIR = scripts
 TEST_DIR = test
 LBIN_DIR = bin
+WORK_DIR = work/tmp
 #if($language == "java")
 TARGET ?= /kb/deployment
 JARS_DIR = $(TARGET)/lib/jars
@@ -120,6 +121,7 @@ build-test-script:
 	echo 'script_dir=$$(dirname "$$(readlink -f "$$0")")' >> $(TEST_DIR)/$(TEST_SCRIPT_NAME)
 	echo 'export KB_DEPLOYMENT_CONFIG=$$script_dir/../deploy.cfg' >> $(TEST_DIR)/$(TEST_SCRIPT_NAME)
 	echo 'export KB_AUTH_TOKEN=`cat /kb/module/work/token`' >> $(TEST_DIR)/$(TEST_SCRIPT_NAME)
+	echo 'rm -rf $(WORK_DIR)' >> $(TEST_DIR)/$(TEST_SCRIPT_NAME)
 #if($language == "perl")
 	echo 'export PERL5LIB=$$script_dir/../$(LIB_DIR):$$PATH:$$PERL5LIB' >> $(TEST_DIR)/$(TEST_SCRIPT_NAME)
 	echo 'cd $$script_dir/../$(TEST_DIR)' >> $(TEST_DIR)/$(TEST_SCRIPT_NAME)
--- a/src/java/us/kbase/templates/module_makefile.vm.properties
+++ b/src/java/us/kbase/templates/module_makefile.vm.properties
@@ -7,7 +7,7 @@ LIB_DIR = lib
 SCRIPTS_DIR = scripts
 TEST_DIR = test
 LBIN_DIR = bin
-WORK_DIR = work/tmp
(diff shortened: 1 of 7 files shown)


MrCreosote:
But will those deleteRecursively() calls work? At this point the java process isn't running as root, right, so I'd guess that they'd fail with a permissions error (or maybe silently depending on the implementation).

MrCreosote:
Maybe that's what the root problem was - kb-sdk test would try and delete the files and fail, so kb-sdk test was blocked. 

That'd be a showstopper if someone was running the sdk remotely on one of our systems, which some people have done.

JamesJeffryes:
I'll try to replicate this on a linux box

MrCreosote:
>  Add a line to test script run in docker container to remove any linge…

That's great until someone writes a test with a 10 GB file...

JamesJeffryes:
@scanon @MrCreosote OK, here's what I settled on. The deleteRecursively() takes care of almost everything (including the subjobs dir) but not nested files in the scratch space on Linux boxes. Those get taken care of by a new command in the test script executed in docker (with root privs) that wipes out anything left in the scratch. Note that for existing repos, these changes will have no effect as the scripts are only generated if missing.

JamesJeffryes:
So yes, if someone, on linux, writes a big file, to a nested subdirectory all that build context will have to go to the docker build but as it was, these files were not even being deleted till after the image was already built

jayrbolton:
The code changes look good to me, but I need to test locally

src/java/us/kbase/templates/module_makefile.vm.properties
```diff
@@ -120,6 +121,7 @@ build-test-script:
 	echo 'script_dir=$$(dirname "$$(readlink -f "$$0")")' >> $(TEST_DIR)/$(TEST_SCRIPT_NAME)
 	echo 'export KB_DEPLOYMENT_CONFIG=$$script_dir/../deploy.cfg' >> $(TEST_DIR)/$(TEST_SCRIPT_NAME)
 	echo 'export KB_AUTH_TOKEN=`cat /kb/module/work/token`' >> $(TEST_DIR)/$(TEST_SCRIPT_NAME)
+	echo 'rm -rf $(WORK_DIR)' >> $(TEST_DIR)/$(TEST_SCRIPT_NAME)
```
jayrbolton: Should this go after the command to run tests? Or do you only want to delete the files before you run tests? I thought maybe it would delete the files after the tests run so the user isn't left with anything sitting in the tmp dir.

jayrbolton: I tried out another variation that adds the following few lines to the end of the `test` target (line 146):
```sh
echo 'Removing temp files...'
rm -rf $(WORK_DIR)
echo '...done removing temp files.'
```

Just so it's very explicit when exactly it happens. And it might be very slightly safter to set `WORK_DIR` to an absolute path like `/kb/module/work/tmp` instead of relative

jayrbolton:
I just had a couple minor comments above. When I make the Makefile modifications, things work well.

I think it's not ideal that this won't work with any existing apps without modifying some of the scripts and the makefile. But I guess that's more of a general downside to the whole templated code approach.. not much we can do about it here.

For the issue of creating/deleting large file sizes, I think many of those large files would be refdata, and it would be nice if refdata was handled by doing a read-only network mount, which would make the cleanup here less of an issue.

src/java/us/kbase/templates/module_makefile.vm.properties
JamesJeffryes: So I've often used the tempfiles for debugging failed tests (or just understanding what kind of output wrapped tools are producing) so I wouldn't want to remove the files after running the tests by default. I could make the the test script a bit more explicit though as you recommend.

JamesJeffryes:
Added in a little more logging per Jay's suggestion. BTW, I looked at rebuilding the makefile on each test run to ensure that these changes get propagated to existing modules but decided against it because I'd like the ability to tweak the makefile in a given repo if I need to.

@scanon Ready for your final review

PR #321, opened 2018-07-18
Sct-1193

JamesJeffryes:
This cleans up the import on the python tests by removing unused import and reordering. Most importantly it removes the dependency on the base image workspace in biokbase so that may be removed at a future date


JamesJeffryes:
@jayrbolton @scanon Quick review on this?

scanon:
Looks good.

PR #323, opened 2018-08-10
Tweak server exception

JamesJeffryes:
If a python3 module is called as a subjob and that module raises an exception a bit of untested code gets called that raises an exception of it's own in python3. The result is a compound exception like this: 
```
  File "/usr/lib/python2.7/dist-packages/nose/suite.py", line 208, in run
    self.setUp()
  File "/usr/lib/python2.7/dist-packages/nose/suite.py", line 291, in setUp
    self.setupContext(ancestor)
  File "/usr/lib/python2.7/dist-packages/nose/suite.py", line 314, in setupContext
    try_run(context, names)
  File "/usr/lib/python2.7/dist-packages/nose/util.py", line 471, in try_run
    return func()
  File "/kb/module/test/phenotype_set_importer_test.py", line 67, in setUpClass
    cls.prepare_data()
  File "/kb/module/test/phenotype_set_importer_test.py", line 92, in prepare_data
    'genome_name': cls.genome_object_name,
  File "/kb/module/lib/GenomeFileUtil/GenomeFileUtilClient.py", line 86, in genbank_to_genome
    job_state = self._check_job(job_id)
  File "/kb/module/lib/GenomeFileUtil/GenomeFileUtilClient.py", line 44, in _check_job
    return self._client._check_job('GenomeFileUtil', job_id)
  File "/kb/module/lib/GenomeFileUtil/baseclient.py", line 216, in _check_job
    return self._call(self.url, service + '._check_job', [job_id])
  File "/kb/module/lib/GenomeFileUtil/baseclient.py", line 183, in _call
    raise ServerError(**err['error'])
ServerError: Unexpected Server Error: 0. An unexpected server error occurred
  File "/kb/module/bin/../lib/GenomeFileUtil/GenomeFileUtilServer.py", line 101, in _call_method
    result = method(ctx, *params)
  File "/kb/module/lib/GenomeFileUtil/GenomeFileUtilImpl.py", line 104, in genbank_to_genome
    result = importer.refactored_import(ctx, params)
  File "/kb/module/lib/GenomeFileUtil/core/GenbankToGenome.py", line 107, in refactored_import
    genome = self.parse_genbank(consolidated_file, params)
  File "/kb/module/lib/GenomeFileUtil/core/GenbankToGenome.py", line 271, in parse_genbank
    self._parse_features(record, params['source'])
  File "/kb/module/lib/GenomeFileUtil/core/GenbankToGenome.py", line 574, in _parse_features
    self.process_cds(_id, feat_seq, in_feature, out_feat)
  File "/kb/module/lib/GenomeFileUtil/core/GenbankToGenome.py", line 761, in process_cds
    raise ValueError(warnings['no_spoof'])
ValueError: Some CDS features in the file do not have a parent gene. Ensure the correct file source is selected, correct the source file or select the 'generate_missing_genes' option.

During handling of the above exception, another exception occurred:

  File "/kb/module/bin/../lib/GenomeFileUtil/GenomeFileUtilServer.py", line 605, in process_async_cli
    resp = application.rpc_service.call_py(ctx, req)
  File "/kb/module/bin/../lib/GenomeFileUtil/GenomeFileUtilServer.py", line 149, in call_py
    respond = self._handle_request(ctx, request)
  File "/kb/module/bin/../lib/GenomeFileUtil/GenomeFileUtilServer.py", line 187, in _handle_request
    result = self._call_method(ctx, request)
  File "/kb/module/bin/../lib/GenomeFileUtil/GenomeFileUtilServer.py", line 118, in _call_method
    if isinstance(e.message, str):
AttributeError: 'ValueError' object has no attribute 'message'
```

The original error is still legible but users could get confused by the additional traceback. This PR fixes this issue by using the args attribute which is available in 2 and 3.

This is a simple fix but will require a rebuild of the python3 image because kb-sdk is a built-in dependency of the base image. (Chicken, meet Egg)This dependency results from the fact that sdk modules specs are compiled by the dockerfile every time.

### DIFF
--- a/src/java/us/kbase/templates/python_server.vm.properties
+++ b/src/java/us/kbase/templates/python_server.vm.properties
@@ -130,9 +130,9 @@ class JSONRPCServiceCustom(JSONRPCService):
             newerr = JSONServerError()
             newerr.trace = traceback.format_exc()
             if len(e.args) == 1:
-                newerr.data = str(e.args[0])
+                newerr.data = repr(e.args[0])
             else:
-                newerr.data = str(e.args)
+                newerr.data = repr(e.args)
             raise newerr
         return result
 
--- a/src/java/us/kbase/templates/python_server.vm.properties
+++ b/src/java/us/kbase/templates/python_server.vm.properties
@@ -129,11 +129,10 @@ class JSONRPCServiceCustom(JSONRPCService):
             # Exception was raised inside the method.
             newerr = JSONServerError()
             newerr.trace = traceback.format_exc()
-            if isinstance(e.message, str):
-                newerr.data = e.message
+            if len(e.args) == 1:
+                newerr.data = str(e.args[0])
             else:
-                # Some exceptions embed other exceptions as the message
-                newerr.data = repr(e.message)
+                newerr.data = str(e.args)
             raise newerr
         return result
 


jayrbolton:
> This is a simple fix but will require a rebuild of the python3 image because kb-sdk is a built-in dependency of the base image. (Chicken, meet Egg)This dependency results from the fact that sdk modules specs are compiled by the dockerfile every time.

:face_with_thermometer:

src/java/us/kbase/templates/python_server.vm.properties
```diff
@@ -129,11 +129,10 @@ class JSONRPCServiceCustom(JSONRPCService):
             # Exception was raised inside the method.
             newerr = JSONServerError()
             newerr.trace = traceback.format_exc()
-            if isinstance(e.message, str):
-                newerr.data = e.message
+            if len(e.args) == 1:
```
jayrbolton: I think this is fine but hacky. I don't know if stdout will log properly in this class, but we could also just do something like:
```py
traceback.print_exc()
print('*' * 80)
print(str(e))
print('*' * 80)
raise JSONServerError()
```

JamesJeffryes:
I think the issue is that narrative try's to display an informative message
in the "error" tab if an exception is raised. If we don't add info to the
"data" attribute the message will always be generic.

jayrbolton:
Okay.. in that case I don't have an alternative. Thumbs up from me. I confirmed locally that `.args` seems to work fine for the old python.

jayrbolton:
If nobody else has comments on this PR, I can merge at the end of the day. 
I don't necessarily want to be the gatekeeper on this repo though... I still don't understand like 85% of the java in here.

JamesJeffryes:
@scanon or @MrCreosote Can we take silence as approval?

src/java/us/kbase/templates/python_server.vm.properties
```diff
@@ -129,11 +129,10 @@ class JSONRPCServiceCustom(JSONRPCService):
...
+                newerr.data = str(e.args[0])
             else:
-                # Some exceptions embed other exceptions as the message
-                newerr.data = repr(e.message)
+                newerr.data = str(e.args)
```
MrCreosote: I specifically remember switching from str to repr because there was info that was being lost with str. I don't remember the specifics though, that was years ago.

src/java/us/kbase/templates/python_server.vm.properties
MrCreosote: Is this backwards compatible with py2 exceptions where message is a field?

MrCreosote: Looks that way
```python
In [1]: v = ValueError('foo')

In [2]: v.message
/usr/local/bin/ipython:1: DeprecationWarning: BaseException.message has been deprecated as of Python 2.6
  #!/usr/bin/python
Out[2]: 'foo'

In [3]: v.args
Out[3]: ('foo',)
```

src/java/us/kbase/templates/python_server.vm.properties
JamesJeffryes: Based on the comment, it was for embedded exceptions. I think all that info still would be present in the traceback though

JamesJeffryes: I think it was actually the error name but you can rest easy. `args` does not have the same issue as message.

```>>> try:
...     func()
... except NameError as e:
...     meh = ValueError(e)
... 
>>> meh.message
NameError("name 'func' is not defined",)
>>> str(meh.message)
"name 'func' is not defined"
>>> repr(meh.message)
'NameError("name \'func\' is not defined",)'
>>> str(meh.args)
'(NameError("name \'func\' is not defined",),)'
>>> repr(meh.args)
'(NameError("name \'func\' is not defined",),)'```

Comment (MrCreosote) [reply src/java/us/kbase/templates/python_server.vm.properties]: `args[0]` does have that issue though:

```python
In [8]: def func():
    def inner():
        raise IOError('io')
    try:
        inner()
    except IOError as e:
        raise ValueError(e)
   ...:     

In [9]: try:     
    func()                  
except ValueError as e:
    err = e
   ...:     

In [10]: str(err)
Out[10]: 'io'

In [11]: repr(err)
Out[11]: "ValueError(IOError('io',),)"

In [12]: str(err.args[0])
Out[12]: 'io'

In [13]: repr(err.args[0])
Out[13]: "IOError('io',)"

In [14]: str(err.args)
Out[14]: "(IOError('io',),)"

In [15]: repr(err.args)
Out[15]: "(IOError('io',),)"
```

JamesJeffryes:
@MrCreosote Thanks for the review. I am now using `repr` for both.

MrCreosote:
I'm concerned that there were no test changes - it sounds like this code isn't covered anywhere, which would've caught the missing info in the revised exception code.

scanon:
To @MrCreosote's comment.  Anyway to add a test?

JamesJeffryes:
@scanon Not an easy way. It's in the templates so can't be tested as standalone code. I tried to get Jay's test environment working but gave up after fighting with VirtualBox a bit.

MrCreosote:
@JamesJeffryes There are standalone tests for the auth client code, so I'm not sure why there couldn't be tests for the baseclient.

MrCreosote:
Looks like there are tests here: https://github.com/kbase/kb_sdk/blob/master/test_scripts/python/test_client.py

although testing intermitted errors might be hard

MrCreosote:
Authclient tests: https://github.com/kbase/kb_sdk/blob/master/test_scripts/py_module_tests/test_auth_client.py

JamesJeffryes:
@MrCreosote This is actually an older issue with the python3 client and involves a tweak to python_server.vm.properties not the baseclient.

MrCreosote:
@JamesJeffryes right, duh, getting my PRs crossed.

That being said there are definitely server tests in the suite; I know because I've spent hours debugging them. The're pretty painful to deal with IMR but it's doable.

JamesJeffryes:
@scanon Do you want to take another look at this one?

PR #324, opened 2018-09-12
Sct-1366: Retry on check_job fail in base client

JamesJeffryes:
This adds logic to retry the _check_job logic if a ConnectionError is raised by requests. This patch seems to work as demonstrated by kb_stringtie and ExpressionUtil modules. 

The other change I am proposing here is to avoid compiling the module on docker build each time. I think it's pretty non intuitive that we commit the compiled code to the repo but that can get out of sync with the actual code run in the docker container because of this step.

### DIFF
--- a/src/java/us/kbase/templates/baseclient.py
+++ b/src/java/us/kbase/templates/baseclient.py
@@ -11,6 +11,8 @@ import json as _json
 import requests as _requests
 import random as _random
 import os as _os
+import traceback as _traceback
+from requests.exceptions import ConnectionError
 
 try:
     from configparser import ConfigParser as _ConfigParser  # py 3
@@ -26,6 +28,7 @@ import time
 _CT = 'content-type'
 _AJ = 'application/json'
 _URL_SCHEME = frozenset(['http', 'https'])
+_CHECK_JOB_RETRYS = 3
 
 
 def _get_token(user_id, password, auth_svc):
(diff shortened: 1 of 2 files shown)


src/java/us/kbase/templates/baseclient.py
```diff
@@ -236,14 +239,22 @@ def run_job(self, service_method, args, service_ver=None, context=None):
...
                 async_job_check_time = self.async_job_check_max_time
-            job_state = self._check_job(mod, job_id)
+
+            try:
+                job_state = self._check_job(mod, job_id)
+            except ConnectionError:
+                _traceback.print_exc()
+                check_job_failures += 1
+                continue
+
```
bio-boris: Maybe want to log that checkjob failed if check_job_failures == _CHECK_JOB_RETRYS

MrCreosote:
I'm pretty sure the catalog expects a compile on build so I don't think the Makefile change is going to work.

src/java/us/kbase/templates/baseclient.py
JamesJeffryes: Actually need to raise and exception at that point. Thx!

JamesJeffryes:
FWIW, you can register and run a python module that does not compile in the dockerfile step. That's how I tested this patch out in stringtie. What happens with Perl or Java and if this is a good idea at all is something I'm willing to discuss

MrCreosote:
So does the documentation in the catalog get built correctly without the compile report? And is the commit in the status() method correct?

JamesJeffryes:
You are right, the catalog service gets the commit directly from the Server file so the compile is currently needed to ensure that that stays current.

JamesJeffryes:
I reverted the change to Makefile as the compile is needed by catalog and that's not going to be trivial to change (having this dependency on kb-sdk in the base image is a major bummer though)

PR #325, opened 2018-10-23
update default method ui spec

psdehal:
pass input mapping by target property and delete the outdated output mapping

### DIFF
--- a/src/java/us/kbase/templates/module_method_spec_json.vm.properties
+++ b/src/java/us/kbase/templates/module_method_spec_json.vm.properties
@@ -145,27 +145,11 @@
             "method": "your_method",
             "input_mapping": [
                 {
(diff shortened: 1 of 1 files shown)


jayrbolton:
So as discussed in person, I think we do want to leave the "workspace" parameter in the default spec for creating reports. I would vote to change it so we eliminate the "target_argument_postion" stuff (have only a single argument with params):

```
            "input_mapping": [
                {
                    "narrative_system_variable": "workspace",
                    "target_property": "workspace_name"
                },
                {
                    "input_parameter": "my_parameter",
                    "target_property": "my_parameter"
                }
            ],

```

JamesJeffryes:
@jayrbolton This PR is superseded by #327

scanon:
Let's close this off and go with #327

PR #327, opened 2018-10-26
SCT-1586: Better default KIDL and UI

JamesJeffryes:
* Make init have default impl that accepts a param dict and returns report refs. 
* Renames "example_module" to "run_<module_name>"
* Updates UI example to modern equivalent.

### DIFF
--- a/src/java/us/kbase/templates/module_method_spec_json.vm.properties
+++ b/src/java/us/kbase/templates/module_method_spec_json.vm.properties
@@ -51,13 +51,14 @@
                 {
                     "narrative_system_variable": "workspace",
                     "target_property": "workspace_name"
-                },
-                {
+                },{
+                    "narrative_system_variable": "workspace_id",
+                    "target_property": "workspace_id"
+                },{
                     "input_parameter": "assembly_input_ref",
                     "target_property": "assembly_input_ref",
                     "target_type_transform": "resolved-ref"
-                },
-                {
+                },{
                     "input_parameter": "min_length",
                     "target_property": "min_length"
                 }
@@ -147,20 +148,21 @@
                 {
                     "narrative_system_variable": "workspace",
                     "target_property": "workspace_name"
-                },
-                {
+                },{
+                    "narrative_system_variable": "workspace_id",
+                    "target_property": "workspace_id"
+                },{
                     "input_parameter": "parameter_1",
                     "target_property": "parameter_1"
                 }
             ],
             "output_mapping": [
                 {
-                "service_method_output_path": [0,"report_name"],
-                "target_property": "report_name"
-                },
-                {
-                "service_method_output_path": [0,"report_ref"],
-                "target_property": "report_ref"
+                    "service_method_output_path": [0,"report_name"],
+                    "target_property": "report_name"
+                },{
+                    "service_method_output_path": [0,"report_ref"],
+                    "target_property": "report_ref"
                 }
             ]
         }
--- a/src/java/us/kbase/templates/module_method_spec_json.vm.properties
+++ b/src/java/us/kbase/templates/module_method_spec_json.vm.properties
@@ -142,29 +142,25 @@
         "service-mapping": {
             "url": "",
             "name": "$module_name",
-            "method": "your_method",
+            "method": "run_$module_name",
             "input_mapping": [
                 {
                     "narrative_system_variable": "workspace",
-                    "target_argument_position": 0
+                    "target_property": "workspace_name"
                 },
                 {
                     "input_parameter": "parameter_1",
-                    "target_argument_position": 1
+                    "target_property": "parameter_1"
                 }
             ],
             "output_mapping": [
                 {
-                    "service_method_output_path": [0],
-                    "target_property": "output"
+                "service_method_output_path": [0,"report_name"],
+                "target_property": "report_name"
                 },
                 {
-                    "input_parameter": "parameter_1",
-                    "target_property": "input"
-                },
-                {
-                    "narrative_system_variable": "workspace",
-                    "target_property": "workspaceName"
+                "service_method_output_path": [0,"report_ref"],
+                "target_property": "report_ref"
                 }
             ]
         }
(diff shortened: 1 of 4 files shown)


src/java/us/kbase/templates/module_typespec.vm.properties
```diff
@@ -84,8 +84,15 @@ module ${module_name} {
...
 #end
 #else
+    typedef structure {
+        string report_name;
+        string report_ref;
+    } ReportResults;
+
     /*
-        Insert your typespec information here.
+        This example function takes accepts any number of parameters and returns results in a KBaseReport
```
jayrbolton: "This example function takes accepts.. " -> "This example function takes.."

src/java/us/kbase/templates/module_method_spec_json.vm.properties
```diff
@@ -142,29 +142,25 @@
         "service-mapping": {
             "url": "",
             "name": "$module_name",
-            "method": "your_method",
+            "method": "run_$module_name",
             "input_mapping": [
                 {
                     "narrative_system_variable": "workspace",
-                    "target_argument_position": 0
+                    "target_property": "workspace_name"
```
MrCreosote: Ideally this would be a workspace ID, not a mutable name.

src/java/us/kbase/templates/module_typespec.vm.properties
```diff
@@ -84,8 +84,15 @@ module ${module_name} {
...
     */
+    funcdef run_${module_name}(UnspecifiedObject params) returns (ReportResults output) authentication required;
```
MrCreosote: Why not make this a mapping<string, UnspecifiedObject> instead? At least then the param names are defined as strings.

src/java/us/kbase/templates/module_method_spec_json.vm.properties
jayrbolton: I think this can be:
```
{
  "target_property": "workspace_id",
  "narrative_system_variable": "workspace_id"
}
```
You can pass this into KBaseReport as long as it's the python version (you added that to the perl version too i think?). See readme on python branch: https://github.com/kbaseapps/KBaseReport/tree/python

JamesJeffryes: I suppose I can switch the example code over to ws_id too

src/java/us/kbase/templates/module_typespec.vm.properties
JamesJeffryes: Sure

src/java/us/kbase/templates/module_method_spec_json.vm.properties
JamesJeffryes: @jayrbolton Did we ever get the python version deployed in prod?

JamesJeffryes: So, AssemblyUtils only takes workspace_name. I'd like these things to be consistent and I think having both would be more confusion that it's worth

jayrbolton: KBaseReport is only in CI still I think.
Have you ever heard of anyone encountering a problem in practical usage with workspace_name, or is it more of a theoretical benefit? If it has never caused a real problem, maybe workspace name is the better choice here. But also it probably wouldn't be hard to update AssemblyUtils

MrCreosote: It's pretty rare that workspace names change (especially for narratives), but it's possible. I'm more concerned that we're putting bad programming practices in our example boilerplate, and devs might continue those bad practices where it does matter more, like in object names.

src/java/us/kbase/templates/module_typespec.vm.properties
JamesJeffryes: $params is a reference to a hash where the key is a string and the value is an UnspecifiedObject, which can hold any non-null object

I do believe that sometimes we get null parameters back so this might prove an issue

JamesJeffryes: Bigger problem, can't use the builder pattern `.withMinLength(-10)` with either form of UnspecifiedObject as the Java example demonstrates.

src/java/us/kbase/templates/module_method_spec_json.vm.properties
JamesJeffryes: So our Utils are a bit of a mess on this. I looked though the APIs to see what would have to change if we wanted consistency.

Only WS ID: GenomeFileUtil, FeatureSetUtil, Generics Util, AssemblyUtil, fba_tools
Both params: KBaseReport, ReadsUtils
Destination ref (effectively both): ReadsAlignmentUtil, DifferentialExpressionUtil, ExpressionUtil
Only WS name: DataFileUtil, TreeUtils

So we might want to have a discussion on what the ideal save API would be. (I'm leaning toward using destination_ref where possible)

src/java/us/kbase/templates/module_typespec.vm.properties
MrCreosote: UObjects can take nulls, so I think that info is out of date:

```
	public static void main(String[] args) throws Exception {
		final ObjectMapper m = new ObjectMapper().registerModule(new JacksonTupleModule());
		
		final String json = m.writeValueAsString(ImmutableMap.of("key", new UObject(null)));
		System.out.println(json);
		
		final Map<String, UObject> instantiated =
				m.readValue(json, new TypeReference<Map<String, UObject>>() {});
		System.out.println(instantiated);
	}
```

MrCreosote: > Bigger problem, can't use the builder pattern .withMinLength(-10) with either form of UnspecifiedObject as the Java example demonstrates.

Yep, the compiler can't compile the fluent setters (not a builder btw, just setters) if it doesn't know the key names and types.

jayrbolton:
We decided in standup to have both `workspace_name` and `workspace_id` as default parameters in the ui spec

JamesJeffryes:
@jayrbolton Updated to pass both workspace name & IDs.

jayrbolton:
:+1:

PR #329, opened 2018-10-31
Add "Hello World" example to basic init

JamesJeffryes:
OK, this was more painful than expected but you should be able to initialize Python, Perl and Java modules and run `kb-sdk test` to produce a hello world report.

### DIFF
--- a/src/java/us/kbase/mobu/initializer/ModuleInitializer.java
+++ b/src/java/us/kbase/mobu/initializer/ModuleInitializer.java
@@ -175,47 +175,43 @@ public class ModuleInitializer {
                 templateFiles.put("module_method_spec_json", Paths.get(moduleDir, "ui", "narrative", "methods", "count_contigs_in_set", "spec.json"));
                 templateFiles.put("module_method_spec_yaml", Paths.get(moduleDir, "ui", "narrative", "methods", "count_contigs_in_set", "display.yaml"));
             }
-			//templateFiles.put("module_test_perl_client", Paths.get(moduleDir, "test", "test_perl_client.pl"));
-			//templateFiles.put("module_test_python_client", Paths.get(moduleDir, "test", "test_python_client.py"));
-			//templateFiles.put("module_test_java_client", Paths.get(moduleDir, "test", "test_java_client.java"));
-			//templateFiles.put("module_test_all_clients", Paths.get(moduleDir, "test", "test_all_clients.sh"));
-
-			switch(this.language) {
-				// Perl just needs an impl file and a start server script
-				case "perl":
-					//templateFiles.put("module_start_perl_server", Paths.get(moduleDir, "scripts", "start_server.sh"));
-				    // start_server script is now made in Makefile
-					templateFiles.put("module_perl_impl", Paths.get(moduleDir, "lib", this.moduleName, this.moduleName + "Impl.pm"));
-					break;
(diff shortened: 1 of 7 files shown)


psdehal:
so the behavior now would be that kb-sdk init creates a run_module function
in the impl file that prints "hello world"

JamesJeffryes:
Actually the created report says "Hello World" (in the test) or what ever you specify in "parameter_1"

jayrbolton:
Testing and looking through things

src/java/us/kbase/mobu/initializer/ModuleInitializer.java
```diff
@@ -175,69 +175,61 @@ public void initialize(boolean example) throws Exception {
...
-						false, // core or sync clients
-						false, // dynamic client
-						"release", //tagVer
-						this.verbose,
-						dependantModuleName,
-						null,
-						null // clientName
-					);
-			}
+            new ClientInstaller(new File(moduleDir), false).install(
```
jayrbolton: lol these params are wild

src/java/us/kbase/templates/module_test_java_client.vm.properties
```diff
@@ -189,6 +192,10 @@ public class ${java_module_name}ServerTest {
         // Check returned data with
         // Assert.assertEquals(..., ret.getSomeProperty());
         // ... or other JUnit methods.
+        Map<String, UObject> params = new HashMap<>();
+        params.put("workspace_name", new UObject(getWsName()));
+        params.put("parameter_1", new UObject("Hello World"));
+        final ReportResults ret = impl.runJavafoo(params, token, getContext());
```
jayrbolton: I think this should be something like `impl.runTestJavaHelloWorld(...`

The java tests don't pass for me as it is now.

jayrbolton: Changing this to `impl.runTestJavaHelloWorld` seems to fix tests

jayrbolton:
There was an issue with the Java test that I commented about above. The perl version seems to work fine (although I can't make heads or tails of that test file, wowzers). The python version throws an error for me:

```py
  File "/miniconda/lib/python3.6/site-packages/nose/failure.py", line 39, in runTest
    raise self.exc_val.with_traceback(self.tb)
  File "/miniconda/lib/python3.6/site-packages/nose/loader.py", line 523, in makeTest
    return self._makeTest(obj, parent)
  File "/miniconda/lib/python3.6/site-packages/nose/loader.py", line 582, in _makeTest
    return MethodTestCase(obj)
  File "/miniconda/lib/python3.6/site-packages/nose/case.py", line 346, in __init__
    self.inst = self.cls()
TypeError: __init__() missing 1 required positional argument: 'config'
```

@JamesJeffryes any ideas? It seems to run the actual test method before the error is raised. I can't tell how or where it gets raised.

JamesJeffryes:
So the Java issue was some bad copy pasta that I missed because I kept using "javafoo" as my module name. I actually can't replicate the Python error you are seeing though. Do you see it when you init with the `-e` flag too?

jayrbolton:
It looks like default tests do work on the filter contigs example, but I still get the above error on the hello world example

jayrbolton:
I finally found the cause of the python error. It's pretty obscure and probably not related to this update.

I was calling my modules "test_python_hello_world_etc". That "test_" prefix on the module name was causing nose2 to do something weird with it. When I rename my module to python_hello_world, it is fixed. Something to note on the troubleshooting page I guess.

PR #338, opened 2018-11-26
Replace "/run" with "/var/run"

jayrbolton:
People are having problems with the latest sdk image and this entrypoint script.

This script hasn't changed for a long time, so I don't know why it started breaking. It must be the new sdk image not having /run. But this should fix it


jayrbolton:
Also, why do we need to mount docker.sock in the first place?

jayrbolton:
I've seen it 3 times now (Sebastian, Bill, and Sam). But I also can't reproduce -- maybe it's because we've monkeyed with our local installs too much.

PR #339, opened 2018-11-28
Update example apps

JamesJeffryes:
Ok, per Shane's request I've changed the example to used the same default "params" mapping as the blank app. This lead to changing the name of the function from `filter_contigs` to `run_${module_name}` for consistency sake. That name change necessitated the bulk of the changes to the sdk tests, template generation code, as well as the language test templates.

For Python and Perl, I chose to extend the "ReportResults" object with the extra counts of total & filtered contigs ect. as this allowed me to leave the existing tests in place. For Java, I had to remove these return values and thus remove those checks.

A final thing added in here is only adding in authclient.py and baseclient.py if a client is being compiled further reducing the files present in the lib/{module} folder.


jayrbolton:
Still working on testing this... Should be able to merge or have feedback shortly after standup.

PR #341, opened 2018-11-29
Cleanup python code

JamesJeffryes:
*merge #339 and #340 to cleanup diff*

This PR is just because once I started on the Python tests I couldn't resist fixing some of the non-idomatic style in the templated impl and tests such as:
- Moves the test object creation into a prepare data function
- Uses `logging` rather than prints so that the logs and test coverage don't mangle one another
- Gets rid of the pointless accessor functions (cuz Python ain't Java)
- Better names for the error test functions


jayrbolton:
If you have successive PRs, each can be a pull request to the branch for the previous PR, rather than them all being into master. That way you have better diffs.

JamesJeffryes:
Brilliant!

PR #342, opened 2019-04-12
fix python3 import

Tianhao-Gu:
@JamesJeffryes @jayrbolton 
Hi guys,
ConfigParser was renamed to configparser in Python3
https://docs.python.org/2/library/configparser.html#module-ConfigParser

### DIFF
--- a/lib/biokbase/log.py
+++ b/lib/biokbase/log.py
@@ -74,7 +74,10 @@ import inspect as _inspect
 import os as _os
 import getpass as _getpass
 import warnings as _warnings
-from ConfigParser import ConfigParser as _ConfigParser
+try:
+    from ConfigParser import ConfigParser as _ConfigParser
+except ImportError:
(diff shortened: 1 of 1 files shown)


JamesJeffryes:
Where are you seeing these files used? I've always thought that they were no longer used

Tianhao-Gu:
@JamesJeffryes hmmm maybe I wasn't doing it in a right way. In workspace test, I have to start a service locally. And I download this file to lib. Because Server.py does `from biokbase import log`

JamesJeffryes:
Yeah, this is confusing. This file is in like 3 places but the one that get's included in the base SDK image (and thus the one Id recommend for you WS docks) is here: https://github.com/kbase/sdkbase2/blob/python/log.py

Tianhao-Gu:
@JamesJeffryes ahh thanks. Closing ticket.

PR #344, opened 2020-01-23
Make python pep8 compliant

eapearson:
The primary functional changes are in the implementation and server template files, simply making the resulting python code pep8 compliant.

The primary change, numerically, is that the template marker comments did not have a space between the comment start (#) and the first non-whitespace character. Empty line rules were the second most numerous offense.
An adjustment was required to to the parser responsible for the impl file so that it can recognize both the old invalid comment markers, as well as properly formatted ones. Without this change the pep8 compliant impl file would trigger an error in kb-sdk.
No tests were added, nor were tests run successfully in preparation of this PR. Tests appear to be either broken or difficult to set up correctly, and I did not have time to address this.
I've used the resulting changes successfully on a couple of projects.

There is more to do.
E.g. src/java/us/kbase/templates/module_python_impl.vm.properties is responsible for generating the initial impl file (and example impl file). The generated impl file is not just pep8 non-compliant but also quite broken and non-functional. Fortunately, running "make" in the module after it is created will rewrite the impl file correctly and install the otherwise missing server file. The module initialization process needs a refactor to ensure that the initial module is good, but that is beyond pep8 compliance and out of scope for the work in this PR.

Also, tests should be fixed or at least fully documented, and test cases adjusted, and pep8/flake8 tests added to ensure the generated code is compliant.


src/java/us/kbase/templates/python_server.vm.properties
```diff
@@ -668,7 +665,3 @@ if __name__ == "__main__":
             assert False, "unhandled option"
 
     start_server(host=host, port=port)
-#    print("Listening on port %s" % port)
-#    httpd = make_server( host, port, application)
-#
-#    httpd.serve_forever()
```
ialarmedalien, line 674: 👍

src/java/us/kbase/templates/python_server.vm.properties
```diff
@@ -146,12 +147,6 @@ class JSONRPCServiceCustom(JSONRPCService):
         debugging purposes.
         """
         rdata = jsondata
-        # we already deserialize the json string earlier in the server code, no
-        # need to do it again
-#        try:
-#            rdata = json.loads(jsondata)
-#        except ValueError:
-#            raise ParseError
```
ialarmedalien, line 154: 👍

eapearson:
@ialarmedalien what do you think about merging this? I was about to fix the .travis config so it would build, but when syncing up with develop saw that it was already fixed, and just pushed that on up in to this branch; now it builds.
I'm starting back up on the ee2 project, which is what instigated this change, since I was sick of my impl file not linting. Of course, this is the develop branch, but there is very little diff between develop and master, so there is hope.

src/java/us/kbase/templates/python_server.vm.properties
eapearson: I thought you'd like the removal of commented-out code :)

ialarmedalien:
This looks good to me. The core change--the regex to search for the parts of the file to replace--should be safe as it searches for specific strings (`# ?(BEGIN|END)_HEADER`, etc.), so there should be no danger of other random comments accidentally matching.

sychan:
Are there any other changes required? This just looks like good hygiene fixes.

ialarmedalien (approved):
LGTM

scanon:
My only concern is making sure existing SDK modules still work and there isn't some odd mismatch behind the scenes.

src/java/us/kbase/templates/python_server.vm.properties
MrCreosote, line 154: A bit of background here - this method is a copy of `jsonrpcbase.call_py` that overrides the original method with a few changes around error handling. I left the commented out code in place so that a later programmer could easily see what changes were made from the parent method and why those changes were made.

PR #345, opened 2020-01-27
Simplifying export declaration and adding strict/warnings

ialarmedalien:
Adding strict and warnings and simplifying the export declaration.

### DIFF
--- a/submodules_hacks/AuthConstants.pm
+++ b/submodules_hacks/AuthConstants.pm
@@ -1,22 +1,34 @@
 package Bio::KBase::AuthConstants;
 
-use constant        globus_token_url   => 'https://nexus.api.globusonline.org/goauth/token?grant_type=client_credentials';
-use constant        globus_profile_url   => 'https://nexus.api.globusonline.org/users';
-use constant	    trust_token_signers => split(/\s+/, 'https://nexus.api.globusonline.org/goauth/keys');
-use constant	    role_service_url => 'https://kbase.us/services/authorization/Roles';
+use strict;
+use warnings;
 
 use base 'Exporter';
-our @EXPORT_OK = qw(globus_token_url globus_profile_url trust_token_signers role_service_url);
-our %EXPORT_TAGS = ( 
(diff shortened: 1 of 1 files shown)


MrCreosote:
this file is totally out of date - none of these urls even exist any more.

MrCreosote:
BTW PRs should be made against the `develop` branch

ialarmedalien:
The main reason for the PR was to check whether there was anything in [Erik's changes](https://github.com/kbase/kb_sdk/pull/344) that might have caused the travis tests to fail. The issue seems to be in [locating the appropriate jars](https://travis-ci.org/kbase/kb_sdk/builds/642619885#L355):

>  [javac] /home/travis/build/kbase/kb_sdk/src/java/us/kbase/catalog/AppClientGroup.java:7: error: package javax.annotation does not exist

Looking at the [PR history in travis](https://travis-ci.org/kbase/kb_sdk/pull_requests), the last successful build was 10 months ago, and had the same travis.yml file. The main (only) difference between that PR and the two most recent PRs is that the `kbase/jars` repo has been updated since then. Presumably there are changes in that repo that are now preventing the travis tests from passing because paths have altered, etc.

MrCreosote:
I doubt it, jars are only added to the repo. I'm guessing that java 8 (maybe 9+? I forget) is now running, which doesn't have the annotation package built in.

In any case, from what I recall the travis tests hardly tested anything, and aren't a good indicator that a set of changes are groovy

MrCreosote:
I looked at `travis.yml`. It barely tests anything

ialarmedalien:
If it's a path / package installation problem, I assume it should be fairly easy to resolve by setting up the appropriate environment.

MrCreosote:
Depends - could wind up peeling the onion as more issues from the java 8 -> java ? switch come to light. 

My advice would be to get the tests running with Java 8, which should fix the annotation package problem, and then see what else, if anything, is broken.

ialarmedalien:
See https://github.com/kbase/kb_sdk/pull/346

PR #346, opened 2020-01-29
Check KBASE_TEST_TOKEN env var if no dev token is found

ialarmedalien:
Two changes in this PR:
- Check for the presence of the environment variable KBASE_TEST_TOKEN if there is no developer token specified in `test_local/test.cfg`
- Specify the JDK version in the travis file config so that the sdk will build successfully on Travis

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,9 @@ language: java
 python:
   - "2.7"
 
+jdk:
+  - openjdk8
+
 sudo: required
 
 services:
@@ -42,4 +45,4 @@ script:
   - make
   - kb-sdk validate
 
-  
+
--- a/src/java/us/kbase/mobu/tester/ConfigLoader.java
+++ b/src/java/us/kbase/mobu/tester/ConfigLoader.java
@@ -29,7 +29,7 @@ public class ConfigLoader {
     private final String catalogUrl;
     private final Map<String, String> secureCfgParams;
 
-    public ConfigLoader(Properties props, boolean testMode, 
+    public ConfigLoader(Properties props, boolean testMode,
             String configPathInfo, boolean tryHomeCfg) throws Exception {
         if (configPathInfo == null) {
             configPathInfo = "test_local/test.cfg";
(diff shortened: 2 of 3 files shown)


.travis.yml
```diff
@@ -3,6 +3,9 @@ language: java
 python:
   - "2.7"
 
+jdk:
+  - openjdk8
```
ialarmedalien, line 7: Ensure that the appropriate JDK is used.

src/java/us/kbase/mobu/tester/ConfigLoader.java
```diff
@@ -47,7 +47,12 @@ public ConfigLoader(Properties props, boolean testMode,
         String tokenString = props.getProperty(testPrefix + "token");
         if (tokenString != null && tokenString.trim().isEmpty()) {
             tokenString = null;
+            String test_token = System.getenv("KBASE_TEST_TOKEN");
+            if (test_token != null && !test_token.trim().isEmpty()) {
+                tokenString = test_token;
+            }
```
ialarmedalien, line 53: Default to using the value in `test.cfg`, and check the env var `KBASE_TEST_TOKEN` if that value is not set.

scanon:
I think you will also need to modify the generated kb-sdk scripts that use Docker so that the variable makes it into the container where the sdk tool runs.

https://github.com/kbase/kb_sdk/blob/master/entrypoint

MrCreosote:
I don't think so, all this happens prior the the entrypoint being called. The token read here winds up in the token file prior to docker being started.

scanon:
What I mean is ...

I set the token variable in my current (non-docker sdk) shell
I envoke kb-sdk which is a script that call docker run .... kb-sdk
   the environment variable in my shell will not be passed into that unless the entrypoint (for the sdk tool) is modified.

MrCreosote:
Oh, right, you're talking about the dockerized version of the SDK. I was thinking of the local version, where this would work as is.

ialarmedalien:
I was intending this for local use, but if you want to add in something to allow it to be used for the dockerised version, go for it.

scanon (approved):

PR #348, opened 2020-06-02
Modify Makefile template to include XML output

jsfillman:
As part of [DEVOPS-247](https://kbase-jira.atlassian.net/browse/DEVOPS-247), I've added XMl output to the Makefile template.

This allows generation & upload of coverage reports as part of our CI build & test process.
This only addresses Python-based apps. Other changes may be needed for Java & Perl app testing.


jsfillman:
Closing, as push to `develop` and then push to Docker via new release is more appropriate path.

PR #350, opened 2020-06-11
Java edits for kb-sdk 1.3

ialarmedalien:
Some changes to prepare kb-sdk for generating GitHub actions and updating the default module builds.

The first commit is all whitespace changes -- my editor removes trailing whitespace and turns tabs into spaces, so I committed all the whitespace changes prior to the content changes. I would suggest viewing just the last three commits.

The changes in this PR comprise:
- adding a build date stamp to `kb-sdk version`
- cleaning up some commented-out lines in the Module Builder
- added in some extra feedback to alert users as to whether auth is using an environment variable or an auth token in a file
- added in a GitHub action that does some rudimentary checks of `kb-sdk` functioning, in lieu of having proper tests.


build.xml
```diff
@@ -102,7 +102,7 @@
   </target>
 
 
-
+  <tstamp/>
```
ialarmedalien, line 105: Timestamp

build.xml
```diff
@@ -116,7 +116,8 @@
     <echo file="${src}/us/kbase/mobu/git.properties">### PLEASE DO NOT CHANGE THIS FILE MANUALLY! ###
 giturl=${git.url}
 branch=${git.branch}
-commit=${git.commit}</echo>
+commit=${git.commit}
+build_timestamp=${DSTAMP}-${TSTAMP}</echo>
```
ialarmedalien, line 120: add date and time to the `git.properties` file

src/java/us/kbase/mobu/ModuleBuilder.java
```diff
@@ -360,7 +360,22 @@ private static int runTestCommand(TestCommandArgs testArgs, JCommander jc) {
...
     private static void printVersion() {
         String gitCommit = getGitCommit();
-        System.out.println("KBase SDK version " + VERSION + (gitCommit == null ? "" : (" (commit " + gitCommit + ")")));
+        String timestamp = getBuildTimestamp();
+        System.out.println("KBase SDK version " + VERSION
+            + (timestamp == null ? "" : ("_" + timestamp))
+            + (gitCommit == null ? "" : (" (commit " + gitCommit + ")")));
+    }
+
+    public static String getBuildTimestamp() {
```
ialarmedalien: Read the build timestamp from the `git.properties` file

src/java/us/kbase/mobu/ModuleBuilder.java
```diff
@@ -360,7 +360,22 @@ private static int runTestCommand(TestCommandArgs testArgs, JCommander jc) {
 
     private static void printVersion() {
         String gitCommit = getGitCommit();
-        System.out.println("KBase SDK version " + VERSION + (gitCommit == null ? "" : (" (commit " + gitCommit + ")")));
+        String timestamp = getBuildTimestamp();
+        System.out.println("KBase SDK version " + VERSION
+            + (timestamp == null ? "" : ("_" + timestamp))
+            + (gitCommit == null ? "" : (" (commit " + gitCommit + ")")));
```
ialarmedalien, line 368: Include the build date and time with the kb-sdk version.

The version string now looks like this:
`KBase SDK version 1.2.0_20200611-1439 (commit 7e58f55ad589fa64f83804c2dc9e380620210c45)` (from GitHub Actions output)

src/java/us/kbase/mobu/initializer/ModuleInitializer.java
```diff
@@ -169,19 +169,16 @@ public void initialize(boolean example) throws Exception {
         }
 
         switch(this.language) {
-            // Perl just needs an impl file and a start server script
+            // start_server scripts are generated by the Makefile
```
ialarmedalien, line 172: clean up comments and remove commented-out lines

src/java/us/kbase/mobu/initializer/ModuleInitializer.java
```diff
@@ -209,38 +204,38 @@ public void initialize(boolean example) throws Exception {
             // Generated examples require some other SDK dependencies
             new ClientInstaller(new File(moduleDir), false).install(
                     this.language,
-                    true, // async clients
-                    false, // core or sync clients
-                    false, // dynamic client
-                    null, //tagVer
+                    true,   // async clients
```
ialarmedalien, line 207: whitespace changes for nicer alignment

src/java/us/kbase/mobu/tester/ConfigLoader.java
```diff
@@ -38,6 +38,11 @@ public ConfigLoader(Properties props, boolean testMode,
         if (authUrl == null)
             throw new IllegalStateException("Error: 'auth_service_url' parameter is not set in " +
                     configPathInfo);
+        endPoint = props.getProperty("kbase_endpoint");
```
ialarmedalien, line 41: moved this check up so that kb-sdk will fail earlier if the `kbase_endpoint` property is not set

src/java/us/kbase/mobu/tester/ConfigLoader.java
```diff
@@ -47,9 +52,13 @@ public ConfigLoader(Properties props, boolean testMode,
         String tokenString = props.getProperty(testPrefix + "token");
         if (tokenString != null && tokenString.trim().isEmpty()) {
             tokenString = null;
+        }
```
ialarmedalien, line 55: adding in some extra messages for help in debugging auth problems (useful when running tests via GitHub actions)

Example output from the GitHub action:
```
No token found in test.cfg file; checking environment
Using token from KBASE_TEST_TOKEN env var
```

src/java/us/kbase/mobu/tester/ModuleTester.java
```diff
@@ -226,9 +226,13 @@ public void logNextLine(String line, boolean isError) {
...
-            int exitCode = ProcessHelper.cmd("bash", DirUtils.getFilePath(runTestsSh),
-                    callbackUrl == null ? "http://fakecallbackurl" :
-                        callbackUrl.toExternalForm()).exec(tlDir).getExitCode();
+            int exitCode = ProcessHelper.cmd(
+                "bash",
+                DirUtils.getFilePath(runTestsSh),
+                callbackUrl == null
+                  ? "http://fakecallbackurl"
+                  : callbackUrl.toExternalForm()
+            ).exec(tlDir).getExitCode();
```
ialarmedalien, line 235: Whitespace changes only -- an upcoming PR has changes to this command and this reformatting makes them easier to understand.

src/java/us/kbase/mobu/util/ProcessHelper.java
```diff
@@ -75,8 +75,9 @@ public ProcessHelper(CommandHolder cmd, File workDir, OutType outType, OutType e
             outPw = output;
         if (error != null)
             errPw = error;
-        process = cmd.cmdLine != null ? Runtime.getRuntime().exec(cmd.cmdLine, null, workDir) :
-                Runtime.getRuntime().exec(cmd.cmdParts, null, workDir);
+        process = cmd.cmdLine != null
+          ? Runtime.getRuntime().exec(cmd.cmdLine, null, workDir)
+          : Runtime.getRuntime().exec(cmd.cmdParts, null, workDir);
```
ialarmedalien, line 80: whitespace changes only.

src/java/us/kbase/mobu/ModuleBuilder.java
```diff
@@ -361,7 +360,22 @@ private static int runTestCommand(TestCommandArgs testArgs, JCommander jc) {
...
+
+    public static String getBuildTimestamp() {
+        String buildTimestamp = null;
+        try {
+            Properties gitProps = new Properties();
+            InputStream is = ModuleBuilder.class.getResourceAsStream("git.properties");
+            gitProps.load(is);
+            is.close();
+            buildTimestamp = gitProps.getProperty("build_timestamp");
+        } catch (Exception ignore) {}
```
MrCreosote: IMO if an exception occurs here the process should fail, as it means the `build.xml` file is borked and thus the build is borked.

Also this is almost completely duplicated by `getGitCommit()` below, DRYification is possible.

.github/workflows/kb_sdk-actions.yml
```diff
@@ -0,0 +1,46 @@
...
+          kb-sdk init -l perl -u user --example PerlExample
+          cd PerlExample
+          kb-sdk test
+          cd ..
+          rm -rf PerlExample
+          kb-sdk init -l python -u user --example MyContigFilter
+          cd MyContigFilter
+          make
+          kb-sdk test
+          cd ..
```
MrCreosote: Why no Java tests? This version of the sdk will probably be used to build Java core services for a long time, so the Java stuff is mission critical.

src/java/us/kbase/mobu/tester/ConfigLoader.java
MrCreosote: This seems to reverse the precedence of where the sdk looks for the token, is that right? That being said, IMO the precedence should be test file > env.

MrCreosote (changes requested):
This *seems* mostly innocuous and it's good to have some automated testing even if it doesn't test most of the code base.

That being said, this is critical infrastructure and making changes without having a real test suite running makes me very nervous.

src/java/us/kbase/mobu/tester/ConfigLoader.java
ialarmedalien: No -- if the token string is not in the test file, the environment will be checked. If KBASE_TEST_TOKEN is non-null and not empty, it will be used.

.github/workflows/kb_sdk-actions.yml
ialarmedalien: I'll put them in. 👍

src/java/us/kbase/mobu/ModuleBuilder.java
ialarmedalien: 👍

src/java/us/kbase/templates/module_java_impl.vm.properties
```diff
@@ -22,7 +22,7 @@ import assemblyutil.GetAssemblyParams;
 import assemblyutil.SaveAssemblyParams;
 import kbasereport.CreateParams;
 import kbasereport.KBaseReportClient;
-import kbasereport.Report;
+import kbasereport.SimpleReport;
```
ialarmedalien, line 25: Java bug fix - incorrect class name.

ialarmedalien:
GitHub Actions results for [commit 9ac1aba](https://github.com/kbase/kb_sdk/commit/9ac1aba13cbf1878be827f5cb44139de997dd85c): https://github.com/ialarmedalien/kb_sdk/actions/runs/137434654

.github/workflows/kb_sdk-actions.yml
```diff
@@ -0,0 +1,136 @@
+name: Basic kb-sdk tests
```
MrCreosote, line 1: FWIW I usually try to do massive whitespace changes in their own PR. Makes the line count for a PR with actual content less scary.

src/java/us/kbase/mobu/ModuleBuilder.java
```diff
@@ -360,24 +361,31 @@ private static int runTestCommand(TestCommandArgs testArgs, JCommander jc) {
...
             Properties gitProps = new Properties();
             InputStream is = ModuleBuilder.class.getResourceAsStream("git.properties");
             gitProps.load(is);
             is.close();
-            gitCommit = gitProps.getProperty("commit");
-        } catch (Exception ignore) {}
-        return gitCommit;
+            propertyValue = gitProps.getProperty(propertyToGet);
+        } catch (Exception e) {
+            showError("Error while retrieving version information", e.getMessage());
```
MrCreosote: `showError` just prints the error, it doesn't exit. I think you'll have to do something else here to stop execution.

.github/workflows/kb_sdk-actions.yml
```diff
@@ -0,0 +1,136 @@
...
+
+jobs:
+
+  test_kb-sdk_builds:
+    if: "!contains(github.event.head_commit.message, 'skip ci')"
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-16.04, ubuntu-18.04, ubuntu-20.04]
```
MrCreosote, line 14: is it possible to test on mac/win? We definitely support mac, and I think we say we support win...?

.github/workflows/kb_sdk-actions.yml
```diff
@@ -0,0 +1,136 @@
...
+          cd SampleApp
+          kb-sdk test || true
+          sed -i "s/test_token=/test_token=$KBASE_TEST_TOKEN/" test_local/test.cfg
+          kb-sdk test
+
+      - name: make resulting app available as artefact in case of failure
+        if: ${{ failure() }}
+        uses: actions/upload-artifact@v2
+        with:
+          name: kbaseapp-${{ matrix.language }}-${{ matrix.test_type }}-${{ matrix.os }}
```
MrCreosote, line 72: I'm guessing each build overwrites the previous build here. Would it make sense to put a timestamp in the name so that doesn't happen, assuming I'm right?

.github/workflows/kb_sdk-actions.yml
```diff
@@ -0,0 +1,136 @@
...
+
+
+  test_existing_repos:
+    if: "!contains(github.event.head_commit.message, 'skip ci')"
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-16.04, ubuntu-18.04, ubuntu-20.04]
+        app: [WsLargeDataIO, KBaseReport]
```
MrCreosote, line 83: Hmm... this means if those repos break then kb-sdk test also breaks. I understand why you're doing this but making the sdk build depend on outside repos building and testing correctly seems overly fragile.

MrCreosote: Also, neither of these repos test callback server functionality

.github/workflows/kb_sdk-actions.yml
MrCreosote, line 14: I suspect many of our machines are still on 12.04 or 14.04... @kkellerlbl?

.github/workflows/kb_sdk-actions.yml
```diff
@@ -0,0 +1,136 @@
...
+      - name: run test
+        run: |
+          cd kbase_app
+          kb-sdk test
+
+      - name: make resulting app available as artefact in case of failure
+        if: ${{ failure() }}
+        uses: actions/upload-artifact@v2
+        with:
+          name: kbaseapp-${{ matrix.app }}-${{ matrix.os }}
```
MrCreosote, line 135: Same ? re timestamp

RELEASE_NOTES.txt
```diff
@@ -1,11 +1,21 @@
...
 
+VERSION: 1.2.2 (Released 06/xx/2020)
+------------------------------------
+NEW FEATURES:
+- kb-sdk version reports the build time and date of the instance
+- kb-sdk can use the environment variable KBASE_TEST_TOKEN if a test token is not present in the test.cfg file
+- GitHub Actions now used to run basic tests on initialising and testing new base and sample repos.
+
+BUG FIXES:
+- Java client uses the correct name for KBase report
```
MrCreosote: note this is for the example module tests

src/java/us/kbase/mobu/tester/ConfigLoader.java
```diff
@@ -47,9 +52,13 @@ public ConfigLoader(Properties props, boolean testMode,
...
+        if (tokenString == null) {
```
MrCreosote, line 56: This still seems to reverse the precedence to my eyes. Old code:
```
ts = getTokenFromConfig();
tt = getTokenFromEnv();
if tt:
  ts = tt
```
New code:
```
ts = getTokenFromConfig();
if not ts:
  tt = getTokenFromEnv();
  ts = tt
```

so in other words old code precedence env > file, new precedence file > env.

Not that that's bad, I think that's the right way to go.

MrCreosote (changes requested):

src/java/us/kbase/mobu/tester/ConfigLoader.java
ialarmedalien: Yes, that's correct. Sorry, I misunderstood what you were saying before.

.github/workflows/kb_sdk-actions.yml
ialarmedalien: This matrix has variables `language`, `test_type`, and `os`, so the name suggested will be unique.

.github/workflows/kb_sdk-actions.yml
ialarmedalien: What would you suggest here? The tests are kind of moot anyway because the app base images may have a different version of kb-sdk on them, so the build/compile step won't really be testing the revised kb-sdk. There should be something to test that the changes aren't going to break existing repos.

.github/workflows/kb_sdk-actions.yml
ialarmedalien: No, only two variables in the matrix, `app` and `os`.

.github/workflows/kb_sdk-actions.yml
ialarmedalien: Unfortunately those systems are too ancient to be testable via GitHub actions!

ialarmedalien: kb-sdk docs explicitly say that we don't support windows -- there are hardcoded paths in various files that mess things up when I tried running the tests on win. On the mac runner, I can't run the `kb-sdk test` step because there's an issue with setting up docker (docker is not available natively on the macOS runner, but is on the ubuntu test runners).

.github/workflows/kb_sdk-actions.yml
ialarmedalien: Yeah, I tried to put them all in the first commit to make it less frightening. Who are these monsters who use tabs and leave trailing spaces all over the place anyway?!

src/java/us/kbase/mobu/ModuleBuilder.java
ialarmedalien: I don't think it's necessary to stop execution if the `git.properties` file can't be found or these properties aren't available -- it's basically just adding in some extra "nice to have" information, it isn't mission critical. The properties file is generated during the build process and deleting it afterwards had no effect on the functioning of kb-sdk.

.github/workflows/kb_sdk-actions.yml
MrCreosote: Only per build, not over time, which is my point. Unless I'm not understanding what you're saying.

.github/workflows/kb_sdk-actions.yml
MrCreosote: Well personally, I wouldn't test against outside repos and I would depend on a good suite of regression tests and quality PR reviews to ensure that changes don't break repos. So let's just make that happen yeah

.github/workflows/kb_sdk-actions.yml
MrCreosote: See response to comment re timestamp above

.github/workflows/kb_sdk-actions.yml
ialarmedalien: Each workflow run has a separate set of artefacts - you download them from the workflow run page. E.g. here's a workflow with failed tests: https://github.com/ialarmedalien/kb_sdk/actions/runs/137590374

MrCreosote: To clarify, I'm talking about the uploads from PRn+1 overwriting the uploads from PRn.

.github/workflows/kb_sdk-actions.yml
ialarmedalien: Hah! I have some ideas for setting up some better regression tests but I didn't want to put too much into this PR, especially since I am working around the knowledge that kb-sdk v1.1.0 to 1.2.0 introduces breaking changes in the python code which have been masked because the sdkbase repo has not been updated. 🤦

I really wish that the old testing mechanisms had not been disabled!

.github/workflows/kb_sdk-actions.yml
ialarmedalien: Every run of the GitHub Actions generates a fresh set of results and artefacts. For example, all the times that I've run GitHub actions on my branches of `kb_sdk`:

https://github.com/ialarmedalien/kb_sdk/actions

src/java/us/kbase/mobu/ModuleBuilder.java
MrCreosote: But why not? If it's not there, clearly something went wrong with the build. Reduce runtime variables IMO.

.github/workflows/kb_sdk-actions.yml
MrCreosote: Were they actually disabled? I just thought that no one could get them to run except me. I'm not even sure if I can get them to run at this point, so I wouldn't be surprised if bitrot had set in.

ialarmedalien: I tried to get them going a while back, but they ran on a VM which has now disappeared, and the config/set up files were removed from the repo.

MrCreosote: > the config/set up files were removed from the repo

That's just... wow.

MrCreosote:
> Java client generated by kb-sdk uses the correct name for KBase report

This still seems off. The Java client doesn't know about the KBase report at all in general. The generated tests for the java module were using an incorrect class name.

ialarmedalien:
> > Java client generated by kb-sdk uses the correct name for KBase report
> 
> This still seems off. The Java client doesn't know about the KBase report at all in general. The generated tests for the java module were using an incorrect class name.

"Corrected a class name in the Java client generated by kb-sdk"?

src/java/us/kbase/mobu/ModuleBuilder.java
ialarmedalien: My thinking is more along the lines of "I don't want to give kb-sdk a reason to die unexpectedly", plus I couldn't create a situation manually where kb-sdk built successfully but the `git.properties` file was not present.

MrCreosote:
I would say 'Corrected a class name in the generated tests for the Java SDK module'. It doesn't really have anything to do with the Java client generally.

If you want to be really specific then 'Corrected a class name for the Java KBaseReport client in the generated tests for the Java SDK module'

src/java/us/kbase/mobu/ModuleBuilder.java
MrCreosote: I'm thinking I don't want unexpected behavior from the SDK when something's broken. I want it to fail fast.

> plus I couldn't create a situation manually where kb-sdk built successfully but the git.properties file was not present.

Then failing won't cause any problems in practice unless something is really wrong, so might as well fail.

.github/workflows/kb_sdk-actions.yml
ialarmedalien: I would like to add the macOS tests -- if/when docker is properly supported via GitHub Actions, it should be very simple just to add it to the actions file.

src/java/us/kbase/mobu/ModuleBuilder.java
ialarmedalien: OK, added a `System.exit(1);` after emitting the error message.

src/java/us/kbase/mobu/ModuleBuilder.java
```diff
@@ -360,24 +361,32 @@ private static int runTestCommand(TestCommandArgs testArgs, JCommander jc) {
...
             InputStream is = ModuleBuilder.class.getResourceAsStream("git.properties");
             gitProps.load(is);
             is.close();
-            gitCommit = gitProps.getProperty("commit");
-        } catch (Exception ignore) {}
-        return gitCommit;
+            propertyValue = gitProps.getProperty(propertyToGet);
+        } catch (Exception e) {
+            showError("Error while retrieving version information from git.properties file: ", e.getMessage());
+            System.exit(1);
```
MrCreosote, line 382: When I was looking at the module builder code earlier, I noticed that most of the methods returned the return code rather than calling exit() themselves. Does that need to happen here? I don't actually understand why they were passing back the return code, but I assume it's for some good reason.

MrCreosote: Maybe testability? If you return the code, you can test the method, but not if you call exit().

ialarmedalien: This happens before any of the module builder methods are called, so you have to just exit. The module builder methods all expect an integer (representing success or otherwise) to be returned:

```java

public static void main(String[] args) throws Exception {
  // code adding the various valid commands and args to JCommander

  // ModuleBuilder line 97: print name and version
  printVersion();

  // parse the arguments and gracefully catch any errors
  try {
    jc.parse(args);
  } catch (RuntimeException e) {
    showError("Command Line Argument Error", e.getMessage());
    System.exit(1);
  }

  // snip some help command stuff

  // if we get here, we have a valid command, so process it and do stuff
  int returnCode = 0;
  if(jc.getParsedCommand().equals(HELP_COMMAND)) {
    showCommandUsage(jc,help,System.out);
  } else if(jc.getParsedCommand().equals(INIT_COMMAND)) {
    returnCode = runInitCommand(initArgs,jc);
  // etc.
```

MrCreosote (approved):
Approving with trepidation as the test suite can't be run.

PR #352, opened 2020-06-22
Update Dockerfile

ialarmedalien:
Edit the dockerfile that `kb-sdk` runs on to update the Docker engine, bump the Java version to Java 8, and bump the Ubuntu release to 20.04.

N.b. the GitHub actions will not run until someone with write access to the repo adds a KBASE_TEST_TOKEN secret under https://github.com/kbase/kb_sdk/settings/secrets

Tests all pass: https://github.com/ialarmedalien/kb_sdk/actions/runs/143963328

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,24 +1,30 @@
-FROM ubuntu:14.04
+FROM ubuntu:20.04
 MAINTAINER Shane Canon <scanon@lbl.gov>
 
 # Update apt and install jdk and docker engine to get docker clients
+# Docker installation instructions from https://docs.docker.com/engine/install/ubuntu/
 RUN apt-get -y update && \
-    apt-get -y install openjdk-7-jdk make git ant && \
-    apt-get -y install apt-transport-https ca-certificates && \
-    apt-key adv \
-               --keyserver hkp://ha.pool.sks-keyservers.net:80 \
-               --recv-keys 58118E89F3A912897C070ADBF76221572C52609D && \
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -1,24 +1,30 @@
...
+# Docker installation instructions from https://docs.docker.com/engine/install/ubuntu/
 RUN apt-get -y update && \
-    apt-get -y install openjdk-7-jdk make git ant && \
-    apt-get -y install apt-transport-https ca-certificates && \
-    apt-key adv \
-               --keyserver hkp://ha.pool.sks-keyservers.net:80 \
-               --recv-keys 58118E89F3A912897C070ADBF76221572C52609D && \
-    echo "deb https://apt.dockerproject.org/repo ubuntu-trusty main" > /etc/apt/sources.list.d/docker.list && \
-    apt-get -y update && apt-get -y install docker-engine=1.11.2-0~trusty 
+    DEBIAN_FRONTEND=noninteractive apt-get -y install tzdata && \
```
ialarmedalien, line 7: `tzdata` asks you to set the time zone (which stops the build process) if not installed with the `non-interactive` option.

Dockerfile
```diff
@@ -1,24 +1,30 @@
...
+    apt-get -y install openjdk-8-jdk make git ant curl gnupg-agent && \
```
ialarmedalien, line 8: `openjdk-7-jdk` is no longer easily available without resorting to all sorts of nasty hackery.

sychan: I don't think we want to have people using jdk7 as a default in 2020

sychan:
The errors are due to no auth token, which is fine since this isn't an app repo.

PR #354, opened 2020-09-10
Adding kb-sdk image build and upload to Dockerhub to GitHub actions

ialarmedalien:
for the data upload project ticket [DATAUP-88](https://kbase-jira.atlassian.net/browse/DATAUP-88)

This adds an automated build and push to Dockerhub (if the appropriate auth credentials are supplied in the repo settings) when merges are made into the `master` and `develop` branches. The dockerhub images will be tagged with `master` and `develop` respectively.

The DockerHub auth hasn't been added to this repo, but here's an image I uploaded to my own dockerhub workspace using the same workflow:

https://hub.docker.com/layers/ialarmedalien/kb-sdk/autobuild_docker_images/images/sha256-8ab997902b5e81c49f0ee9bade1f278cfbb2446191da2b658cd8a303d48305b6?context=repo

### DIFF
--- a/.github/workflows/kb_sdk-actions.yml
+++ b/.github/workflows/kb_sdk-actions.yml
@@ -134,3 +134,33 @@ jobs:
         with:
           name: kbaseapp-${{ matrix.app }}-${{ matrix.os }}
           path: kbase_app
+
+  dockerhub_upload:
+    runs-on: ubuntu-latest
+    if: |
+      (github.ref == 'refs/heads/develop' || github.ref == 'refs/heads/master') &&
+      github.event_name == 'push' &&
+      !contains(github.event.head_commit.message, 'skip docker') &&
+      !contains(github.event.head_commit.message, 'skip ci')
+    needs:
+      - test_existing_repos
+      - test_kb-sdk_builds
+
+    steps:
+    - name: checkout git repo
+      uses: actions/checkout@v2
+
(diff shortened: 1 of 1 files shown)


jsfillman:
One minor suggestion I'd make is to set the branch triggers directly vs. in a conditional.
Using a conditional will cause the workflow to _always_  be triggered, but skip steps based on the condition. Setting  `branches` will cause the whole workflow to be skipped.

In the following example, the workflow will only trigger on PRs to develop or master, and only when the PR is created, switched from a draft review, or when it receives new commits.


```
---
name: Build Image
'on':
  pull_request:
    branches:
    - master
   - develop
    types:
    - opened
    - synchronize
    - ready_for_review
```

Feel free to steal from the narrative(-traefiker) workflow:
https://github.com/kbase/narrative-traefiker/blob/master/.github/workflows/build_prodrc_pr.yaml

ialarmedalien:
I'm using the conditional because the other workflow steps in the file run on every pull request and push to any branch; the build/dockerhub step only needs to run when there's a merge to develop or master.

jsfillman (approved):

jayrbolton (approved):
The sdk is released so rarely (like once every 18 months?) that i think automating the deployment isn't very important, but it is nice to standardize the github workflow alongisde our other repos.

PR #356, opened 2020-09-18
Better perl

ialarmedalien:
Part I of some improvements I made a while back to the perl sdk app templates. I'm using the fully improved perl sdk app for the Jacobson / Cytoscape app (kb_cytoscape).

[Commit 1](https://github.com/kbase/kb_sdk/commit/420bc6b7146cdecec32585b42fd309a5aa9482f6) is almost entirely whitespace changes from running the code formatter `perltidy` over the codebase.

[Commit 2](https://github.com/kbase/kb_sdk/commit/48286f8c62e6186c4ddff7aefdfa84f1600efd00) has more simple changes, including:
- some more perltidy tidy up
- adding `use strict` and `use warnings` to all files so errors get reported
- use the correct form of class initialisation, which is `Class::Name->new`, not `new Class::Name`
- removed `$PATH` from the `$PERL5LIB` env var -- `$PATH` is for executables, `$PERL5LIB` is for libraries
- a few minor variable renamings

Note: the tests are marked as failed because this is from my own forked repo. They have passed on my repo: https://github.com/ialarmedalien/kb_sdk/tree/better_perl (note little green tick).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,4 @@
 /test_scripts/test.cfg
 *.pyc
 env-*/
+submodules/auth
--- a/lib/Bio/KBase/Log.pm
+++ b/lib/Bio/KBase/Log.pm
@@ -431,7 +431,7 @@ sub _syslog {
 sub _log {
     my ( $self, $ident, $message ) = @_;
     my $time = POSIX::strftime( "%Y-%m-%d %H:%M:%S", localtime );
-    my $msg  = join( " ", $time, hostname(), $ident . ": " );
+    my $msg  = join " ", $time, hostname(), $ident . ": ";
     open LOG, ">>" . $self->get_log_file()
         || warn "Could not print log message $msg to " . $self->get_log_file() . "\n";
     if ( ref( $message ) eq 'ARRAY' ) {
--- a/lib/Bio/KBase/Log.pm
+++ b/lib/Bio/KBase/Log.pm
@@ -14,51 +14,52 @@ use Time::HiRes;
 use Sys::Syslog qw( :DEFAULT setlogsock);
 
 # $ENV{'MLOG_CONFIG_FILE'} should point to an INI-formatted file, or an empty string, or should not exist.
-our $MLOG_ENV_FILE = "MLOG_CONFIG_FILE";
-my $_GLOBAL = "global";
-our $MLOG_LOG_LEVEL = "mlog_log_level";
-our $MLOG_API_URL = "mlog_api_url";
-our $MLOG_LOG_FILE = "mlog_log_file";
+our $MLOG_ENV_FILE      = "MLOG_CONFIG_FILE";
+my $_GLOBAL             = "global";
+our $MLOG_LOG_LEVEL     = "mlog_log_level";
+our $MLOG_API_URL       = "mlog_api_url";
+our $MLOG_LOG_FILE      = "mlog_log_file";
 
-our $DEFAULT_LOG_LEVEL = 6;
-our $MSG_FACILITY = 'local1';
-our $EMERG_FACILITY = 'local0';
+our $DEFAULT_LOG_LEVEL  = 6;
+our $MSG_FACILITY       = 'local1';
+our $EMERG_FACILITY     = 'local0';
 
-our $EMERG = 0;
-our $ALERT = 1;
-our $CRIT = 2;
-our $ERR = 3;
+our $EMERG   = 0;
+our $ALERT   = 1;
+our $CRIT    = 2;
+our $ERR     = 3;
 our $WARNING = 4;
-our $NOTICE = 5;
-our $INFO = 6;
-our $DEBUG = 7;
-our $DEBUG2 = 8;
-our $DEBUG3 = 9;
+our $NOTICE  = 5;
+our $INFO    = 6;
+our $DEBUG   = 7;
+our $DEBUG2  = 8;
+our $DEBUG3  = 9;
 
-my $_MLOG_TEXT_TO_LEVEL = { 'EMERG' => $EMERG,
-                            'ALERT' => $ALERT,
-                            'CRIT' => $CRIT,
-                            'ERR' => $ERR,
-                            'WARNING' => $WARNING,
-                            'NOTICE' => $NOTICE,
-                            'INFO' => $INFO,
-                            'DEBUG' => $DEBUG,
-                            'DEBUG2' => $DEBUG2,
-                            'DEBUG3' => $DEBUG3,
-                            };
+my $_MLOG_TEXT_TO_LEVEL = {
+    'EMERG'   => $EMERG,
+    'ALERT'   => $ALERT,
+    'CRIT'    => $CRIT,
+    'ERR'     => $ERR,
+    'WARNING' => $WARNING,
+    'NOTICE'  => $NOTICE,
+    'INFO'    => $INFO,
+    'DEBUG'   => $DEBUG,
+    'DEBUG2'  => $DEBUG2,
+    'DEBUG3'  => $DEBUG3,
+};
 
-my @_MLOG_TO_SYSLOG = (0, 1, 2, 3, 4, 5, 6, 7, 7, 7);
+my @_MLOG_TO_SYSLOG = ( 0, 1, 2, 3, 4, 5, 6, 7, 7, 7 );
 
 my $_MLOG_LEVEL_TO_TEXT = {};
-foreach my $k (keys %{$_MLOG_TEXT_TO_LEVEL}) {
-        $_MLOG_LEVEL_TO_TEXT->{$_MLOG_TEXT_TO_LEVEL->{$k}} = $k;
+foreach my $k ( keys %{ $_MLOG_TEXT_TO_LEVEL } ) {
+    $_MLOG_LEVEL_TO_TEXT->{ $_MLOG_TEXT_TO_LEVEL->{ $k } } = $k;
 }
 
-our $LOG_LEVEL_MIN = min(keys %{$_MLOG_LEVEL_TO_TEXT});
-our $LOG_LEVEL_MAX = max(keys %{$_MLOG_LEVEL_TO_TEXT});
+our $LOG_LEVEL_MIN = min( keys %{ $_MLOG_LEVEL_TO_TEXT } );
+our $LOG_LEVEL_MAX = max( keys %{ $_MLOG_LEVEL_TO_TEXT } );
 
-our $USER = $ENV{'USER'};
-our $PARENT_FILE = abs_path($0);
+our $USER        = $ENV{ 'USER' };
+our $PARENT_FILE = abs_path( $0 );
 
 =pod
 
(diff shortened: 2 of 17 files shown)


src/java/us/kbase/templates/module_dockerfile.vm.properties
```diff
@@ -3,7 +3,6 @@ FROM kbase/sdkbase2:python
 #else
 FROM kbase/sdkbase2:latest
 #end
-MAINTAINER #if($username)${username}#{else}KBase Developer#{end}
```
ialarmedalien, line 6: This tag has been deprecated

src/java/us/kbase/templates/module_dockerfile.vm.properties
```diff
@@ -34,15 +33,13 @@ RUN R -q -e 'if(!require(raster)) install.packages("raster", repos="http://cran.
...
 COPY ./ /kb/module
-RUN mkdir -p /kb/module/work
-RUN chmod -R a+rw /kb/module
 
 WORKDIR /kb/module
 
-RUN make all
+RUN mkdir -p /kb/module/work && \
+    chmod -R a+rw /kb/module && \
+    make all
```
ialarmedalien, line 42: combine all directives into a single statement

src/java/us/kbase/templates/module_perl_impl.vm.properties
```diff
@@ -215,47 +247,17 @@ Return the module status. This is a structure including Semantic Versioning numb
...
+    return ( $return );
 }
 
 1;
-#else
-#BEGIN_HEADER
-use Bio::KBase::AuthToken;
-use installed_clients::KBaseReportClient;
-use Config::IniFiles;
-#END_HEADER
```
ialarmedalien, line 232: Some of this is duplication of what's in the sample app, and unless you run `make`, the app that gets produced by default is just a bunch of random stanzas. I've integrated this and the sample app code together so that both the base app and the sample app are valid, sensible code without needing to run `make` / invoking the SDK.

src/java/us/kbase/templates/perl_client.vm.properties
```diff
@@ -115,67 +115,84 @@ sub new
...
     bless $self, $class;
-    #    $self->_validate_version();
     return $self;
 }
 
 #foreach( $module in $modules )
 #if( $any_async || $async_version)
-sub _check_job {
-    my($self, @args) = @_;
 # Authentication: ${method.authentication}
```
ialarmedalien, line 144: moved the `# Authentication` line out of the method and into the header

src/java/us/kbase/templates/perl_client.vm.properties
```diff
@@ -115,67 +115,84 @@ sub new
...
-        if ($result->is_error) {
-            Bio::KBase::Exceptions::JSONRPC->throw(error => $result->error_message,
-                           code => $result->content->{error}->{code},
-                           method_name => '_check_job',
-                           data => $result->content->{error}->{error} # JSON::RPC::ReturnObject only supports JSONRPC 1.1 or 1.O
-                          );
+
+    my ( $job_id ) = @args;
+
+    my @bad_arguments;
```
ialarmedalien, line 157: `@_bad_arguments` doesn't need that underscore, especially since it gets reported to the user and is therefore not private!

MrCreosote:
Sorry, I don't understand why the tests would fail in your repo but pass here...? Is it due to secrets not working on PRs?

lib/Bio/KBase/Log.pm
```diff
@@ -174,275 +175,303 @@ clear_user_log_level() : Removes the user-defined log level.
...
-    if(defined $self->{_mlog_config_file} && -e $self->{_mlog_config_file} &&
-             -s $self->{_mlog_config_file} > 0) {
-        my $cfg = new Config::Simple($self->{_mlog_config_file});
-        my $cfgitems = $cfg->get_block($_GLOBAL);
-        my $subitems = $cfg->get_block($self->{_subsystem});
-        foreach my $k (keys %{$subitems}) {
-            $cfgitems->{$k} = $subitems->{$k};
+    if ( defined $self->{ _mlog_config_file }
+        && -e $self->{ _mlog_config_file }
+        && -s $self->{ _mlog_config_file } > 0 ) {
```
MrCreosote, line 266: Having the conditional and the inner block have the same level of indentation makes it pretty hard to read quickly.

lib/Bio/KBase/Log.pm
```diff
@@ -174,275 +175,303 @@ clear_user_log_level() : Removes the user-defined log level.
...
 }
 
 sub _resolve_log_level {
-    my ($self, $level) = @_;
-    if(defined $_MLOG_TEXT_TO_LEVEL->{$level}) {
-        $level = $_MLOG_TEXT_TO_LEVEL->{$level};
-    } elsif (!(defined $_MLOG_LEVEL_TO_TEXT->{$level})) {
+    my ( $self, $level ) = @_;
+    if ( defined $_MLOG_TEXT_TO_LEVEL->{ $level } ) {
+        $level = $_MLOG_TEXT_TO_LEVEL->{ $level };
```
MrCreosote, line 335: I personally strongly dislike whitespace between brackets / parentheses and contents, but *shrug*

lib/Bio/KBase/Log.pm
```diff
@@ -174,275 +175,303 @@ clear_user_log_level() : Removes the user-defined log level.
...
-        return $self->{_user_log_level};
-    } elsif($self->{_config_log_level} != -1) {
-        return $self->{_config_log_level};
-    } elsif($self->{_api_log_level} != -1) {
-        return $self->{_api_log_level};
-    } else {
+    my ( $self ) = @_;
+    if ( $self->{ _user_log_level } != -1 ) {
+        return $self->{ _user_log_level };
+    }
```
MrCreosote, line 241: putting the `elsif` on its own like makes it harder to read, as it's less clear that it's a continuation of the prior `if`.

lib/Bio/KBase/Log.pm
```diff
@@ -174,275 +175,303 @@ clear_user_log_level() : Removes the user-defined log level.
...
+    elsif ( $self->{ _config_log_level } != -1 ) {
+        return $self->{ _config_log_level };
+    }
+    elsif ( $self->{ _api_log_level } != -1 ) {
+        return $self->{ _api_log_level };
+    }
+    else {
```
MrCreosote, line 248: Same for `else`.

lib/Bio/KBase/Log.pm
MrCreosote: Multiple places where this occurs in this file.

lib/Bio/KBase/Log.pm
MrCreosote: Multiple places where this occurs in this file.

lib/Bio/KBase/Log.pm
```diff
@@ -174,275 +175,303 @@ clear_user_log_level() : Removes the user-defined log level.
...
+    $self->{ _callback }();
 }
 
 sub set_log_msg_check_count {
-    my ($self, $count) = @_;
-    if($count !~ /^\d+$/ || $count < 0) {
-        die "Format for calling set_log_msg_check_count is set_log_msg_check_count(integer count)\n";
+    my ( $self, $count ) = @_;
+    if ( $count !~ /^\d+$/ || $count < 0 ) {
+        die
```
MrCreosote, line 369: Why would the die message go on a separate line?

MrCreosote: This also occurs in multiple places.

src/java/us/kbase/templates/module_dockerfile.vm.properties
MrCreosote: I would want @sychan to weigh in on this one since he created all the hooks/scripts etc that create and use these variables and he arguably best understands how they interact. Seems innocuous but worth double checking.

src/java/us/kbase/templates/module_dockerfile.vm.properties
```diff
@@ -34,15 +33,13 @@ RUN R -q -e 'if(!require(raster)) install.packages("raster", repos="http://cran.
 RUN apt-get -y install r-cran-evaluate r-cran-codetools r-cran-testthat
 #end
 
-# -----------------------------------------
```
MrCreosote, line 37: Seems like this shouldn't be deleted, it's supposed to separate the part of the file the user can edit from the part they shouldn't

ialarmedalien:
@MrCreosote yes, exactly.

src/java/us/kbase/templates/module_dockerfile.vm.properties
scanon: We should replace it with LABEL maintainer="...."

src/java/us/kbase/templates/module_perl_impl.vm.properties
MrCreosote: Thanks for this explanation, the changes threw me for a loop at first

src/java/us/kbase/templates/module_perl_impl.vm.properties
```diff
@@ -65,130 +67,160 @@ sub new
...
-    my($output);
+    my ( $output );
     #BEGIN run_${module_name}
-    
+#if ($example)
     # Print statements to stdout/stderr are captured and available as the App log
-    print("Starting run_${module_name} method. Parameters:\n");
-    print(Dumper($params) . "\n");
-    
+    say "Starting run_${module_name} method. Parameters: " . Dumper $params;
```
MrCreosote, line 99: Should there be a newline before `Dumper $params;`?

src/java/us/kbase/templates/perl_client.vm.properties
MrCreosote: Sorry, I don't understand this change. Can you further explain what's going on here?

src/java/us/kbase/templates/perl_client.vm.properties
MrCreosote: Can you even access variables inside a function scope?

src/java/us/kbase/templates/perl_client.vm.properties
```diff
@@ -572,69 +677,66 @@ sub call {
...
-	}
-#if( $enable_client_retry )
+        if ($uri =~ /\?/) {
+            $result = $self->_get($uri);
+        }
+        else {
+            Carp::croak "not hashref." unless (ref $obj eq 'HASH');
+            $result = $self->_post($uri, $headers, $obj);
+        }
+        #if( $enable_client_retry )
```
MrCreosote, line 695: Should this be indented?

src/java/us/kbase/templates/perl_client.vm.properties
```diff
@@ -572,69 +677,66 @@ sub call {
...
-	    {
-		$want_retry = 1;
-	    }
-	    
-	}
-	
+        my $msg = $result->message;
+        my $want_retry = 0;
+        if ($codes_to_retry${empty_escaper}{$code})
+        {
```
MrCreosote, line 717: this isn't following the pattern of brace on `if` line

.gitignore
```diff
@@ -11,3 +11,4 @@
 /test_scripts/test.cfg
```
MrCreosote, line 11: What proportion of changes in this PR are covered by tests?

src/java/us/kbase/templates/module_dockerfile.vm.properties
ialarmedalien: Done.

src/java/us/kbase/templates/perl_impl.vm.properties
```diff
@@ -44,12 +45,11 @@ ${module.module_constructor}    #END_CONSTRUCTOR
 #set( $status_in_kidl = true )
 #end
 
-
 =head2 ${method.name}
 
   #if( "$method.ret_vars" != "" )${method.ret_vars} = #{end}$obj->${method.name}(${method.arg_vars})
 
-=over 4
+=over
```
MrCreosote, line 52: Why the deletion? I'm not particularly familiar with perldoc

sychan (commented):
This review seems to have stalled. Given that there are few people who have experience with the Perl functionality beyond AJ, I'm going to merge this by COB today unless I see a substantive criticism.

MrCreosote:
I'm still working on it. There's just a lot of changes and I've had a lot of meetings the past couple days.

src/java/us/kbase/templates/perl_server.vm.properties
```diff
@@ -38,455 +42,505 @@ our %return_counts = (
...
-sub _build_config
-{
-    my ($self) = @_;
-    my $sn = $self->get_service_name();
-    my $cf = $self->get_config_file();
-    if (!($cf)) {
+sub _build_config {
+    my ( $self )    = @_;
+    my $sn          = $self->get_service_name;
+    my $cf          = $self->get_config_file;
```
MrCreosote, line 101: I'm not a perl expert, but this looks like to me that the vars will be assigned the functions themselves, rather than the return value of the functions. If this is some sort of odd syntax where the functions get called even without the parens is there a reason to do it this way? It's very confusing to someone who doesn't know the trick.

src/java/us/kbase/templates/perl_server.vm.properties
```diff
@@ -38,455 +42,505 @@ our %return_counts = (
...
-    my ($self, $error) = @_;
-    
-    unless (ref($error) eq 'HASH' ||
-           (blessed $error and $error->isa('RPC::Any::Exception'))) {
-        $error = RPC::Any::Exception::PerlError->new(message => $error);
+    my ( $self, $error ) = @_;
+
+    unless ( ref( $error ) eq 'HASH'
+        || ( blessed $error and $error->isa( 'RPC::Any::Exception' ) ) ) {
+        $error = RPC::Any::Exception::PerlError->new( message => $error );
```
MrCreosote, line 174: Having the same indentation for the conditional and the block is difficult to read

src/java/us/kbase/templates/perl_server.vm.properties
```diff
@@ -503,274 +557,242 @@ is available via $context->client_ip.
...
+
+    if ( $dest && $dest =~ m!^/! ) {
+        #
+        # File destination
+        #
+        my $fh;
+        if ($tag) {
+            $tag =~ s,/,_,g;
+            $dest = "$dest/$tag";
+            mkdir($dest) if ! -d $dest;
```
MrCreosote, line 667: This line is actually swapping the logic around a bit rather than just whitespace - is this tested somehow?

MrCreosote: Normally I wouldn't worry about it but the tests for kb_sdk don't run right now so I'm on the wary side

src/java/us/kbase/templates/perl_server.vm.properties
```diff
@@ -503,274 +557,242 @@ is available via $context->client_ip.
...
-    my($self) = @_;
-    if ($self->{dest})
-    {
-	return("2>", $self->{dest});
-    }
-    else
-    {
-	return ();
-    }
+sub redirect {
```
MrCreosote, line 694: The logic here is also shuffled around, so same question re testing

src/java/us/kbase/templates/perl_server.vm.properties
```diff
@@ -503,274 +557,242 @@ is available via $context->client_ip.
...
-    my($self) = @_;
-    if ($self->{dest})
-    {
-	return(">&", $self->{dest});
-    }
-    else
-    {
-	return ();
-    }
+sub redirect_both {
```
MrCreosote, line 702: logic shuffling

src/java/us/kbase/templates/perl_server.vm.properties
```diff
@@ -503,274 +557,242 @@ is available via $context->client_ip.
...
-    else
-    {
-	return $self->{file};
+sub text_value {
+    my ( $self ) = @_;
+    if ( ref( $self->{ dest } ) eq 'SCALAR' ) {
+        my $r = $self->{ dest };
+        return $$r;
     }
+    return $self->{ file };
```
MrCreosote, line 764: Also logic shuffling here, but pretty minor logic shuffling

src/java/us/kbase/templates/module_dockerfile.vm.properties
MrCreosote: That's interesting, review comments are added to the comment thread based on the day they're made (although the date still shows `now`) rather than being added to the end of the comment thread as one would expect

src/java/us/kbase/templates/perl_client.vm.properties
ialarmedalien: prior to change:
```
sub _check_job {
    my ($self, @args) = @_;
    # Authentication: ...
```

after change:
```
# Authentication
sub _check_job {
    my ($self, @args) = @_;
```

i.e. the comment isn't stuck in the middle of the method

src/java/us/kbase/templates/module_perl_impl.vm.properties
ialarmedalien: No need.

src/java/us/kbase/templates/perl_client.vm.properties
ialarmedalien: You can't access `@bad_arguments` outside `_check_job`, no.

src/java/us/kbase/templates/perl_client.vm.properties
ialarmedalien: The indentation of this whole section is messed up -- there seems to be a lot of confusion on the part of the author on when new scopes need to be created... well, just general confusion about perl. I have addressed this in a future commit.

The line `#if( $enable_client_retry )` is a template directive, so its indentation does not matter too much as it won't appear in the output.

src/java/us/kbase/templates/perl_client.vm.properties
ialarmedalien: The perltidy linter didn't catch this, probably because it's got some template variables in the line (the `$codes_to_retry${empty_escaper}{$code}`). I have gone over a fair amount of this code by hand, so it's likely to be fixed in a future PR.

src/java/us/kbase/templates/perl_server.vm.properties
ialarmedalien: You don't have to use brackets when calling a function in perl; if there are no arguments, calling `$self->do_something` will do the same as `$self->do_something()`. There are a lot of functions that can be called with or without brackets -- e.g. `print`, `join`, `split`, `die`, etc., etc. Generally perl programmers will omit the brackets unless they are required to disambiguate an expression.

src/java/us/kbase/templates/perl_server.vm.properties
ialarmedalien: I just used the default perltidy setting.

src/java/us/kbase/templates/perl_server.vm.properties
ialarmedalien: In perl, you can put conditionals after an expression, e.g.

```perl
do_this( $args ) if $condition;
```
or
```perl
$var = $value unless defined $var;
```

It means you can write a lot of these types of conditional expressions in a single line, instead of having to put them in a `if (...) {  }` stanza.

src/java/us/kbase/templates/perl_server.vm.properties
ialarmedalien: Ditto explanation from above.

src/java/us/kbase/templates/perl_server.vm.properties
ialarmedalien: Not logic shuffling, just removing an unneeded `else`. If the `if` ends with a `return`, there's no need to put in an `else`.

src/java/us/kbase/templates/module_perl_impl.vm.properties
ialarmedalien: It threw me for even more of a loop when I first generated a perl kbase app! At least now there's a sensible impl file returned after init-ing a repo.

lib/Bio/KBase/Log.pm
ialarmedalien: I'm the opposite - I find it's easier to read code when things are more spread out.

lib/Bio/KBase/Log.pm
ialarmedalien: this is the default perltidy setting

lib/Bio/KBase/Log.pm
ialarmedalien: Maybe to keep the whole string intact? I don't know. I didn't look over the whitespace in every file in detail for this PR, this was just to impose standard formatting on all the perl files before looking over the code itself.

.gitignore
ialarmedalien: The vast majority of the changes are generated by perltidy, which is a popular and very commonly-used code formatter:

https://metacpan.org/pod/distribution/Perl-Tidy/bin/perltidy

This is the first in a series of updates to the perl templates that I have made and have been using on my own repo. I have added in tests for perl file compilation and rewritten the test script to use the perl test harness properly (I have also retrofitted those changes to a couple of the existing perl kbase apps -- e.g. RAST SDK; this means test failures will be reported, rather than just dying silently). This PR just applies some standardised formatting.

src/java/us/kbase/templates/module_dockerfile.vm.properties
ialarmedalien: Docker docs on MAINTAINER tag: https://docs.docker.com/engine/reference/builder/#maintainer-deprecated

sychan: The PR is correct, the MAINTAINER directive has been obsolete for a while and newer image should use LABEL MAINTAINER=

src/java/us/kbase/templates/perl_server.vm.properties
sychan: Yes, this is idiomatic Perl.

src/java/us/kbase/templates/perl_server.vm.properties
sychan: This just idiomatic perl.

lib/Bio/KBase/Log.pm
sychan: This is a personal stylistic preference.

sychan (approved):
The majority of these changes are automatic formatting changes that perltidy applied, and while we may not personally like the styling, having a standard formatter is a huge step in the right direction.

The change to the Dockerfile was a legitimate question, but the change was actually correct, because the MAINTAINER directive became obsolete after the original dockerfile was generated.

Unless some more substantive issues come up, I think this PR is ok to merge.

PR #357, opened 2020-09-22
Remove return statement that stops kb-sdk test

ialarmedalien:
Currently, if you run `kb-sdk test` for the first time, it will exit after creating the `test.cfg` configuration file. However, since `kb-sdk` can now extract dev tokens from the environment (specifically from the env var `KBASE_TEST_TOKEN`), it is possible for the test run to go ahead as `test.cfg` and the environment may contain all the information needed to run the tests. If there is no token in the environment or `test.cfg`, `kb-sdk test` will die with the usual message about "Error: KBase account credentials are not set...".

To test this, I have removed the steps from the GitHub actions where `test.cfg` is edited to add the test token, and just populated the environment variable instead. The tests now run without requiring the extra `kb-sdk test || true` step.

I have also edited the error message about missing KBase account credentials to state that they were not found in either `test.cfg` or the env var `KBASE_TEST_TOKEN`.

### DIFF
--- a/.github/workflows/kb_sdk-actions.yml
+++ b/.github/workflows/kb_sdk-actions.yml
@@ -61,8 +61,6 @@ jobs:
           LANGUAGE_TOKEN: ${{ matrix.language }}
         run: |
           cd SampleApp
-          kb-sdk test || true
-          sed -i "s/test_token=/test_token=$KBASE_TEST_TOKEN/" test_local/test.cfg
           kb-sdk test
 
       - name: make resulting app available as artefact in case of failure
@@ -111,19 +109,9 @@ jobs:
           repository: kbaseapps/${{ matrix.app }}
           path: kbase_app
 
-      - name: prep for repo tests
-        run: |
-          cd kbase_app
-          kb-sdk test || true
-
-      - name: edit test.cfg
+      - name: run repo tests
         env:
           KBASE_TEST_TOKEN: ${{ secrets.KBASE_TEST_TOKEN }}
-        run: |
-          cd kbase_app
-          sed -i "s/test_token=/test_token=$KBASE_TEST_TOKEN/" test_local/test.cfg
-
-      - name: run test
         run: |
           cd kbase_app
           kb-sdk test
--- a/.github/workflows/kb_sdk-actions.yml
+++ b/.github/workflows/kb_sdk-actions.yml
@@ -95,6 +95,7 @@ jobs:
       matrix:
         os: [ubuntu-16.04, ubuntu-18.04, ubuntu-20.04]
         app: [WsLargeDataIO, KBaseReport]
+        auth: [token, env_var]
 
     steps:
       - name: Checkout
@@ -125,7 +126,23 @@ jobs:
           repository: kbaseapps/${{ matrix.app }}
           path: kbase_app
 
-      - name: run repo tests
+      - name: run repo tests, using test.cfg for auth
+        if: matrix.auth == 'token'
+        run: |
+          cd kbase_app
+          kb-sdk test || true
+
+      - name: run repo tests, using test.cfg for auth
+        if: matrix.auth == 'token'
+        env:
+          KBASE_TEST_TOKEN: ${{ secrets.KBASE_TEST_TOKEN }}
+        run: |
+          cd kbase_app
+          sed -i "s/test_token=/test_token=$KBASE_TEST_TOKEN/" test_local/test.cfg
+          kb-sdk test
+
+      - name: run repo tests, using env var for auth
+        if: matrix.auth == 'env_var'
         env:
           KBASE_TEST_TOKEN: ${{ secrets.KBASE_TEST_TOKEN }}
         run: |
--- a/.github/workflows/kb_sdk-actions.yml
+++ b/.github/workflows/kb_sdk-actions.yml
@@ -14,6 +14,7 @@ jobs:
         os: [ubuntu-16.04, ubuntu-18.04, ubuntu-20.04]
         language: [perl, python, java]
         test_type: [base, example]
+        auth: [token, env_var]
 
     steps:
       - name: Checkout
@@ -55,10 +56,25 @@ jobs:
         run: |
           kb-sdk init -l $LANGUAGE_TOKEN -u user --example SampleApp
 
-      - name: test ${{ matrix.test_type }} repo
+      - name: test ${{ matrix.test_type }} repo, using test.cfg for auth
+        if: matrix.auth == 'token'
+        run: |
+          cd SampleApp
+          kb-sdk test || true
+
+      - name: test ${{ matrix.test_type }} repo, using test.cfg for auth
+        if: matrix.auth == 'token'
         env:
           KBASE_TEST_TOKEN: ${{ secrets.KBASE_TEST_TOKEN }}
-          LANGUAGE_TOKEN: ${{ matrix.language }}
+        run: |
+          cd SampleApp
+          sed -i "s/test_token=/test_token=$KBASE_TEST_TOKEN/" test_local/test.cfg
+          kb-sdk test
+
+      - name: test ${{ matrix.test_type }} repo, using env var for auth
+        if: matrix.auth == 'env_var'
+        env:
+          KBASE_TEST_TOKEN: ${{ secrets.KBASE_TEST_TOKEN }}
         run: |
           cd SampleApp
           kb-sdk test
--- a/src/java/us/kbase/mobu/tester/ConfigLoader.java
+++ b/src/java/us/kbase/mobu/tester/ConfigLoader.java
@@ -64,7 +64,7 @@ public class ConfigLoader {
 
         if (user == null && tokenString == null) {
             throw new IllegalStateException("Error: KBase account credentials are not set in " +
-                    configPathInfo);
+                    configPathInfo + " or the KBASE_TEST_TOKEN environment variable");
         }
         authAllowInsecure = props.getProperty("auth_service_url_allow_insecure");
         ConfigurableAuthService auth = new ConfigurableAuthService(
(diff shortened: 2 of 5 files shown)


sychan (commented):
This seems like a reasonable change.

src/java/us/kbase/mobu/tester/ConfigLoader.java
```diff
@@ -64,7 +64,7 @@ public ConfigLoader(Properties props, boolean testMode,
 
         if (user == null && tokenString == null) {
             throw new IllegalStateException("Error: KBase account credentials are not set in " +
-                    configPathInfo);
+                    configPathInfo + " or the KBASE_TEST_TOKEN environment variable");
```
MrCreosote, line 67: probably should make KBTT a static var

.github/workflows/kb_sdk-actions.yml
```diff
@@ -111,19 +109,9 @@ jobs:
           repository: kbaseapps/${{ matrix.app }}
           path: kbase_app
 
-      - name: prep for repo tests
```
MrCreosote, line 114: Shouldn't a test where the test.cfg file is exercised be included?

.github/workflows/kb_sdk-actions.yml
```diff
@@ -61,8 +61,6 @@ jobs:
           LANGUAGE_TOKEN: ${{ matrix.language }}
```
MrCreosote: A general concern I have is that this makes it easier for users to mismatch their token and environment, which happens all the time even with a test file where both token and environment are in the same context. Now the token will be in an env  var and the environment in the test.cfg file.

ialarmedalien: There's already a barrier to entry for using this code, which is that users have to set an environment variable; not all users will know how to do this, and will probably be put off by that requirement.

For users who are comfortable with it, having dev tokens defined as env vars in their shell can make kbase life quicker and easier; e.g. I have `KBASE_TEST_TOKEN_APPDEV`, `KBASE_TEST_TOKEN_CI`, and `KBASE_TEST_TOKEN` set up in my bashrc, and I alter the last of those to whichever value is appropriate for what I'm doing. It's also very convenient for automated testing via GitHub actions.

Ideally it would be possible to set the server to test against (appdev / CI / etc.) via env var, too, to facilitate automated testing against different servers without having to resort to hacky `sed` statements to edit files in place.

.github/workflows/kb_sdk-actions.yml
ialarmedalien: Done.

.github/workflows/kb_sdk-actions.yml
MrCreosote: Hmm, if very few users are going to figure out how to use this feature, is it worth the maintenance burden?

sychan: The incremental amount of code is quite small, and it cleans up how things runs. I support it.

sychan: This doesn't add another other opportunities for user confusion about their environments, so while that is a general concern, it is applies to all such configurations.

.github/workflows/kb_sdk-actions.yml
```diff
@@ -55,16 +56,29 @@ jobs:
...
-      - name: test ${{ matrix.test_type }} repo
+      - name: test ${{ matrix.test_type }} repo, using test.cfg for auth
+        if: matrix.auth == 'token'
+        run: |
+          cd SampleApp
+          kb-sdk test || true
+
+      - name: test ${{ matrix.test_type }} repo, using test.cfg for auth
+        if: matrix.auth == 'token'
         env:
```
MrCreosote, line 67: Does this env line need to be here given it's a token in file test?

.github/workflows/kb_sdk-actions.yml
```diff
@@ -111,19 +125,9 @@ jobs:
           repository: kbaseapps/${{ matrix.app }}
           path: kbase_app
 
-      - name: prep for repo tests
-        run: |
```
MrCreosote, line 115: I guessing you think that running both env and file tests for the app tests is overkill?

.github/workflows/kb_sdk-actions.yml
MrCreosote: > The incremental amount of code is quite small

Note there's more code for this change in other PRs.

MrCreosote: > This doesn't add another other opportunities for user confusion about their environments

I don't think this is true. As I pointed out in the OC, it's a lot easier to figure out when you have a mismatch when the token and env specification live in the same place.

.github/workflows/kb_sdk-actions.yml
ialarmedalien: Yes -- otherwise the `sed` command will not be able to add the token to the test.cfg.

src/java/us/kbase/mobu/tester/ConfigLoader.java
sychan: Are you talking about the tokenString var? Wouldn't this same thing apply to almost all of the other strings declared in this method like password, testPrefix, user, authUrl, etc...?

.github/workflows/kb_sdk-actions.yml
ialarmedalien: I can add it in. IMO it's somewhat redundant with the other set of tests that generate new apps, which exercise both env var and test.cfg testing.

src/java/us/kbase/mobu/tester/ConfigLoader.java
MrCreosote: I mean `KBASE_TEST_TOKEN` is probably present as a string all over the place and it should probably be a static variable somewhere, and the places where it's used should use that static var instead.  Worst case put it in a `Constants` class or something like that.

I'm sure there's plenty of other cases like this that duplicate strings, but this is the one that shows up in the diff so I'm mentioning it :P 

No reason to continue a bad practice just because it's already there, etc

.github/workflows/kb_sdk-actions.yml
MrCreosote: If you don't think it's necessary I'm fine with that; I just want to understand the reason for the change.

MrCreosote: I suppose another alternative is running env tests with one app and file tests with the other, but I don't feel strongly about it

ialarmedalien: I'll just switch to the same paradigm as the other tests, checking both the env var and the file tests.

src/java/us/kbase/mobu/tester/ConfigLoader.java
sychan: KBASE_TEST_TOKEN is a shell environment variable though, isn't it? If you are talking about the text literal in the code on line 58, that seems to be the only place it occurs in Java code:
https://github.com/kbase/kb_sdk/search?q=KBASE_TEST_TOKEN

MrCreosote: How does the code get the token from the environment if KBTT is only used in an error message?

MrCreosote: https://github.com/kbase/kb_sdk/blob/8fc4c6041cd27518476e96b928cc02c290063d73/src/java/us/kbase/mobu/tester/ConfigLoader.java#L58

MrCreosote: One other spot in the same file

sychan: My bad, I should have been more clear: it is only used to fetch the env var once in the Java code across the entire repo, on line 58 as I mentioned. Otherwise it is just as part of output string and not used anywhere else in code. So we're really only talking about a single instance of KBASE_TEST_TOKEN (it is not all over the place as originally feared). The only place a static var declaration would apply would be in this file, and only for that one time that it is referenced.

MrCreosote: There's 3 total places the string literal occurs in the code

MrCreosote: The location we're talking about now doesn't show up in search

MrCreosote: We're spending a lot of time discussing a pretty trivial change...

sychan: Now at least we know it is a trivial change, and not a sprawling one. I only see only one place that KBASE_TEST_TOKEN is evaluated, and the concern that it is all over the place and needs to be a static declaration doesn't apply.

sychan: I don't see a strong case for making it a static declaration, but if AJ feels like making the change, I'm sure nobody would object!

ialarmedalien: This doesn't seem to me like a worthwhile change to make, given where and how the string is used in that file. As you say, a lot of time is being spent discussing a trivial change; if it means a lot to @MrCreosote, I'm happy to review a PR about it.

MrCreosote: Duplicate string literals is a pretty standard code smell: 
http://rc3.org/2006/10/11/use-contants-instead-of-string-literals/
https://rules.sonarsource.com/java/RSPEC-1192
https://sqlenlight.com/help_19/html/c04d63ab-e82f-41e4-b55f-9a7852a25211.htm

MrCreosote: > Now at least we know it is a trivial change, and not a sprawling one.

Even if there were dozens of instances of the string it'd still be a trivial change.

sychan (approved):

PR #359, opened 2020-10-09
Update entrypoint script to force root user

bio-boris:
Since the new mac osx update, kb-sdk no longer properly works. The user inside the kb-sdk container doesn't have permissions to the docker socket. 

* Possible side effects are having to use `sudo su` to  to delete the files that appear on your machine


ialarmedalien:
Is the issue with OS X or with Docker? I heard conflicting reports about what was causing the problem.

ialarmedalien (approved):
👍

PR #361, opened 2020-10-15
Fix Merge Conflicts 

bio-boris:
Fix Merge Conflicts in order to merge to master. Both dev and master are protected


bio-boris:
?!?! is master just a broken branch? and we should override the entire src/java/us/kbase/mobu/ModuleBuilder.java  file??

PR #362, opened 2020-10-22
Removing "-t" flag to allow github actions to run kb-sdk

ialarmedalien:
GitHub Actions cannot run kb-sdk when it is in a docker container because the `entrypoint` script adds the `-t` flag to the docker command (to add a TTY-type interface to the container), which causes GitHub Actions to error out. Since the commands that the entry point script is adding can't be run interactively, the `-t` flag can be safely removed.

### DIFF
--- a/RELEASE_NOTES.txt
+++ b/RELEASE_NOTES.txt
@@ -6,6 +6,10 @@ platform.  Modules include all code, specification files, and
 documentation needed to define and run a set of methods
 in the KBase Narrative interface.
 
+VERSION: 1.2.6 (Released 10/2020)
+---------------------------------
+CHANGES:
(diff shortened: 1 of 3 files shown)


briehl:
I had to workaround this on some Github actions stuff by editing the `genscript` results. This would be great to merge before the next image release.

scanon:
LGTM

jsfillman (approved):

jsfillman:
This eliminates the need for a custom `kb-sdk` command, like I've been using in the [SDK Tests Action](https://github.com/marketplace/actions/kbase-sdk-tests) as seen [here](https://github.com/kbaseapps/kb_sdk_actions/blob/master/bin/kb-sdk). Excellent!

Issue #347, opened 2020-03-20
kb-sdk v1.2: python tests fail


ialarmedalien:
Currently the Docker `kbase/sdkbase2:python` base image has v1.1 of `kb-sdk` installed on it. If you try to update `kb-sdk` to v1.2, running `kb-sdk test` fails as follows:

```
Running /home/runner/work/kb_sdk/kb_sdk/MyContigFilter/test_local/run_docker.sh with args run -v /home/runner/work/kb_sdk/kb_sdk/MyContigFilter/test_local/workdir:/kb/module/work -e SDK_CALLBACK_URL=http://172.17.0.1:38903 -e KB_SDK_TEST=1 test/mycontigfilter:latest test

Run Tests
if [ ! -f /kb/module/work/token ]; then echo -e '\nOutside a docker container please run "kb-sdk test" rather than "make test"\n' && exit 1; fi
bash test/run_tests.sh
Running test/run_tests.sh with args 
Removing temp files...
Finished removing temp files.
E
======================================================================
ERROR: Failure: ModuleNotFoundError (No module named 'MyContigFilter.authclient')
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/miniconda/lib/python3.6/site-packages/nose/failure.py", line 39, in runTest
    raise self.exc_val.with_traceback(self.tb)
  File "/miniconda/lib/python3.6/site-packages/nose/loader.py", line 418, in loadTestsFromName
    addr.filename, addr.module)
  File "/miniconda/lib/python3.6/site-packages/nose/importer.py", line 47, in importFromPath
    return self.importFromDir(dir_path, fqname)
  File "/miniconda/lib/python3.6/site-packages/nose/importer.py", line 94, in importFromDir
    mod = load_module(part_fqname, fh, filename, desc)
  File "/miniconda/lib/python3.6/imp.py", line 235, in load_module
    return load_source(name, filename, file)
  File "/miniconda/lib/python3.6/imp.py", line 172, in load_source
    module = _load(spec)
  File "<frozen importlib._bootstrap>", line 684, in _load
  File "<frozen importlib._bootstrap>", line 665, in _load_unlocked
  File "<frozen importlib._bootstrap_external>", line 678, in exec_module
  File "<frozen importlib._bootstrap>", line 219, in _call_with_frames_removed
  File "/kb/module/test/MyContigFilter_server_test.py", line 8, in <module>
    from MyContigFilter.MyContigFilterServer import MethodContext
[...]

PR #367, opened 2022-06-22
Update module_test_python_client.vm.properties

bio-boris:
See #347 
Called at https://cs.github.com/kbase/kb_sdk/blob/07a0a3ce72060510b58202ce8151f4d4818a81aa/src/java/us/kbase/mobu/initializer/ModuleInitializer.java#L152
and
https://cs.github.com/kbase/kb_sdk/blob/07a0a3ce72060510b58202ce8151f4d4818a81aa/src/java/us/kbase/mobu/compiler/TemplateBasedGenerator.java#L206

### DIFF
--- a/src/java/us/kbase/templates/module_test_python_client.vm.properties
+++ b/src/java/us/kbase/templates/module_test_python_client.vm.properties
@@ -6,7 +6,11 @@ from configparser import ConfigParser
 
 from ${module_name}.${module_name}Impl import ${module_name}
 from ${module_name}.${module_name}Server import MethodContext
-from ${module_name}.authclient import KBaseAuth as _KBaseAuth
+
+try:
+    from ${module_name}.authclient import KBaseAuth as _KBaseAuth
+except ImportError:
+    from installed_clients.authclient import KBaseAuth as _KBaseAuth
 
 #if ($example)
 from installed_clients.AssemblyUtilClient import AssemblyUtil
--- a/src/java/us/kbase/templates/python_server.vm.properties
+++ b/src/java/us/kbase/templates/python_server.vm.properties
@@ -18,7 +18,10 @@ from jsonrpcbase import ServerError as JSONServerError
 
 from biokbase import log
 #if( $authenticated )
-from ${modules[0].pypackage}authclient import KBaseAuth as _KBaseAuth
+try:
+    from ${modules[0].pypackage}authclient import KBaseAuth as _KBaseAuth
+except ImportError:
+    from installed_clients.authclient import KBaseAuth as _KBaseAuth   
 #end
 
 try:


MrCreosote:
Why are there almost 100 checks? That seems insane

MrCreosote:
Making changes to kb_sdk makes me very nervous due to the centrality of the module in KBase, its complexity, and the fact that the tests are unrunnable (except *maybe* on my magic VM, but IIRC they weren't working the last time I tried, which was years ago). 

I assume that whatever change broke things, and caused this change to be necessary, got through because of the lack of tests.

The change itself looks reasonable at face value. The questions I have are
* How and where did this bug get introduced?
* What testing has been done to make sure the change works and is backwards compatible?

bio-boris:
1) I did not investigate how and when this bug was introduced
2) As for testing, I did it late at night and should probably redo it, but the testing I did was
* built the image with `docker build .  -t kbase-sdk:0.2.6-dev` and updated my local `/bin/kb-sdk` to use this new built sdk
* deleted the modules server file, and ran make to regenerate it and see if it generated the tests correctly

* TODO is to create a new module using this new kb-sdk
* TODO: Play with this             initPythonPackages(pythonServerPath, output, false); to see if that's the issue or not?

* As for being backwards compatible, I think it will NOT be backwards compatible with this fix, but all versions since the bug introduced and have been "released" (but not pushed to dockerhub) have this issue as well..
 The way to may it backwards compatible will be to in addition to this fix, make the kb-sdk copy the clients into the correct location and/or update the makefile to correctly generate a testing script with the right PYTHONPATH.. 

(I'm not entirely sure if its possible to change paths and move things around and would need to look into further. Making these changes in the MAKEFILE is not a portable solution as people may have updated the makefile..)

MrCreosote:
Can we figure out how the bug came to be? That'll also help out re figuring out whether the change is backward compatible or not. It'd be bad if we needed to recompile an old module and it breaks

bio-boris:
Looking at the git blame, the bug came to be when it got added to the template

https://github.com/kbase/kb_sdk/commit/3fe3e8875bccc2d97ddcca802fede3ff00f506b0#diff-c4bd14f7c977da591d617fdf8e484fd558391135ad1607871c93b549cff53875R19

Hmm, the other one got added https://github.com/kbase/kb_sdk/blob/3fe3e8875bccc2d97ddcca802fede3ff00f506b0/src/java/us/kbase/templates/module_test_python_client.vm.properties way back though so maybe it was during the python2-3 update? Not sure.

I think these changes actually make it more backwards compatible.

Wait this wouldn't affect recompiling old modules, because these steps only happen during init of the module I believe.

MrCreosote:
It looks to me that whenever the change was made to put all the clients into `installed_clients` the templates weren't updated correctly and still pointed to the old locations. I think both the commits above are prior to that

MrCreosote:
One of the files is the python server template; I'm pretty sure that gets recompiled (retemplated?) every time make compile is run

MrCreosote:
For testing you say you should probably rerun the tests, maybe when you do that you could try recompiling a really old module like DFU and seeing if it still works, recompiling a newer module (not sure what's new) and seeing it it still works, and making a new module and checking it works, along with the tests you outlined above

ialarmedalien (approved):
I have been annoyed by this bug for several years now!

ialarmedalien:
I was looking at the "insane" GHA test failures in [another kb-sdk PR](https://github.com/kbase/kb_sdk/pull/370). Some of the failures are due to GHA issues (timeouts, etc.), but there are also problems with the example module that gets created as part of the GHA tests, specifically that the tool it installed (something from sourceforge) is no longer available.

PR #372, opened 2024-04-12
Pass Token Via File

bio-boris:
* Pass token via file rather than the command line args, so you cannot see the token with `ps`
* Related https://github.com/kbaseapps/kbase_hipmer/pull/19


MrCreosote:
Unfortunately this will cause all Java apps to break, probably cause all Perl apps to break, and I presume cause python apps compiled before the token file feature was added to break

bio-boris:
Is this file used for generating templates in `kb-sdk init`? If so, wouldn't it mean that new apps would generate the async file this way and not affect existing python modules? I'm putting this here as a possible thing to include we can add in the kb-sdk:python fork. I will close this and use an issue instead.

MrCreosote:
> Is this file used for generating templates in kb-sdk init? If so, wouldn't it mean that new apps would generate the async file this way and not affect existing python modules?

I'm not sure if it's init or compile, probably init, in which case only new java & perl apps would break

